PR #15, opened 2019-06-05
add hide to main file settings

v-kents:


Odata-docs/docfx.json
```diff
@@ -3,8 +3,8 @@
     "content": [
       {
         "files": [
-          "**/*.md",
-          "**/*.yml"
+          "**/*.{md,yml}",
```
jdanyow: this line should be `"*.{md,yml}"`

v-kents: I think I had it that way a few versions back, but changing to see

v-kents:
merging to see if a full build fixes remaining files

Issue #53, opened 2019-09-04
leverage IODataClientFactory to Use HttpClient in OData Client


chuanboz:
in [Use HttpClient in OData Client](https://docs.microsoft.com/en-us/odata/client/using-httpclient), it talked about how to use HttpClient in OData Client, and includes lots of sample code to make that work. With the addtion to ODataClient extensions (https://github.com/OData/Extensions/), there is new way to use HttpClient in ODataClient, very similar to IHttpClientFactory experience and would work out of box.

The issues is to add public document for that out of box usage and link this to that page.
@xuzhg FYI.

PR #67, opened 2019-10-16
Add client extensions change log and usage document, plus sample for Unit Test

chuanboz:
This also addressed issue #53


madansr7:
@chuanboz , the changes look good. However i don't seen any additions to the ToC.yml file so not sure how this is going to be discovered.

chuanboz:
@madansr7 , thanks for the help review it, I've added change to TOC as you suggested.

PR #72, opened 2019-10-25
merging from master

madansr7:


Rick-Anderson:
@madansr7  can you merge this so the content will be live? I'd like to start on my next ASP.NET Core ODATA tutorial but these need to be live first.

PR #92, opened 2020-01-07
Remove white-spaces

ruant:


ruant:
Looks like there is a lot of documents that's missing author.

PR #111, opened 2020-02-27
Add documentation for dsc when using POCO's

marabooy:
The PR introduces a new action to load the metadata from the metadata Url and set the ServiceModel. The new action runs to load the metadata if the metadata had not been set before (directly by calling context.Format.UseJson(ServiceModel) or using a user provided Hook context.Format.LoadServiceModel.
This would allow people to use the syntax below to use this method in their code when not using the code generation tools or when they want to use the latest model in their code. E.g when developing new API's or when working on API's that do not publish breaking changes.

https://github.com/OData/odata.net/pull/1614

### DIFF
--- a/Odata-docs/client/v6/basic-crud-operations.md
+++ b/Odata-docs/client/v6/basic-crud-operations.md
@@ -12,6 +12,10 @@ ms.topic: article
 
 **Applies To**: [!INCLUDE[appliesto-odataclient](../../includes/appliesto-odataclient-v6.md)]
 
+## With Codegen/Connected service
+The odata client can be used by generating code using the odata connected service or odata codegen. You can learn more on this  from this [tutorial](https://docs.microsoft.com/en-us/aspnet/web-api/overview/odata-support-in-aspnet-web-api/odata-v4/create-an-odata-v4-client-app "Create an OData v4 Client App (C#)") 
+
+The code
 ## Request an entity set
 
 ```csharp
@@ -20,7 +24,7 @@ var context = new DefaultContainer(new Uri("https://services.odata.org/v4/(S(lqb
 var people = context.People.Execute();
 ```
 
-The `Execute()` API call will return an `IEnumerable<Person>`.
+The `Execute()` API call will return an `IEnumerable<Person>`. To use the async api the call for `Execute` can be replaced with `ExecuteAsync` which returns a `Task<IEnumerable<Person>>`
 
 ## Request an individual entity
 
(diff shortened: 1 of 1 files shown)


Odata-docs/client/v6/basic-crud-operations.md
```diff
@@ -20,7 +24,7 @@ var context = new DefaultContainer(new Uri("https://services.odata.org/v4/(S(lqb
 var people = context.People.Execute();
 ```
 
-The `Execute()` API call will return an `IEnumerable<Person>`.
+The `Execute()` API call will return an `IEnumerable<Person>`. To use the async api the call for `Execute` can be replaced with `ExecuteAsync` which returns a `Task<IEnumerable<Person>>`
```
odero: Instead of discussing the returned type in the docs it might be easier to just specify the type in the code snippet instead of using `var`

Odata-docs/client/v6/basic-crud-operations.md
```diff
@@ -74,3 +78,64 @@ context.DeleteObject(person); // create a delete request
...
+    public Container(Uri serviceRoot) : base(serviceRoot)
+    {
+              this.People = base.CreateQuery<Person>("People");
+    }
+
+    public DataServiceQuery<Person> People {get; }
+
+    public void AddToPeople(Person person)
+    {
+       base.AddObject("People",person);
```
odero:
```suggestion
       base.AddObject("People", person);
```

Odata-docs/client/v6/basic-crud-operations.md
```diff
@@ -74,3 +78,64 @@ context.DeleteObject(person); // create a delete request
...
+
+// Dataservice context
+ public class Container : DataServiceContext
+{
+    public Container(Uri serviceRoot) : base(serviceRoot)
+    {
+              this.People = base.CreateQuery<Person>("People");
+    }
+
+    public DataServiceQuery<Person> People {get; }
```
odero:
```suggestion
    public DataServiceQuery<Person> People { get; }
```

Odata-docs/client/v6/basic-crud-operations.md
```diff
@@ -74,3 +78,64 @@ context.DeleteObject(person); // create a delete request
 
 context.SaveChanges(); // send the request
 ```
+
+
+## Using with POCO's
+
+While developing you may have the need to use the client without using codegen so as to reduce duplication of code for the models. In this case one can still use the odata client
```
odero:
```suggestion
While developing, you may want to use OData Client without a code generation tool like _ODataConnectedService_. In this case, one can still use the OData Client
```

Odata-docs/client/v6/basic-crud-operations.md
```diff
@@ -74,3 +78,64 @@ context.DeleteObject(person); // create a delete request
 
 context.SaveChanges(); // send the request
 ```
+
+
+## Using with POCO's
+
+While developing you may have the need to use the client without using codegen so as to reduce duplication of code for the models. In this case one can still use the odata client 
+to access their service.
```
odero: this line is dangling

Odata-docs/client/v6/basic-crud-operations.md
odero: Also if you could fix the following lines:
L60, L76: `UserName` --> `userName`
L67: Please be noted --> Please note

Odata-docs/client/v6/basic-crud-operations.md
```diff
@@ -12,6 +12,10 @@ ms.topic: article
 
 **Applies To**: [!INCLUDE[appliesto-odataclient](../../includes/appliesto-odataclient-v6.md)]
 
+## With Codegen/Connected service
```
odero:
```suggestion
## With Connected Service
```
Also this heading should be at different level than "Request an entity set" or "Request an individual entity"

odero: Considering you are introducing a new way of doing things you need a better intro for entire "Basic Crud Operations". It should clearly indicate that there are 2 ways: using ConnService and using POCO.
Are there any benefits or risks for choosing one approach over the other? This should be clearly outlined
In which use cases should one be chosen over the other? (Should also be outlined)

Odata-docs/client/v6/basic-crud-operations.md
```diff
@@ -12,6 +12,10 @@ ms.topic: article
...
+The odata client can be used by generating code using the odata connected service or odata codegen. You can learn more on this  from this [tutorial](https://docs.microsoft.com/en-us/aspnet/web-api/overview/odata-support-in-aspnet-web-api/odata-v4/create-an-odata-v4-client-app "Create an OData v4 Client App (C#)") 
+
+The code
```
odero: This seems out of place. You could have a header for "Examples" instead

Odata-docs/client/v6/basic-crud-operations.md
```diff
@@ -74,3 +78,64 @@ context.DeleteObject(person); // create a delete request
...
+    {
+       base.AddObject("People",person);
+    }
+}
+
+```
+
+The above context can then be used the same way as the codegen data service context. 
+
+To add new objects to the peoples entity set use `context.AddObject("People",person)` or create a convenience method in the class to do the same.
```
odero:
```suggestion
To add new objects to the peoples entity set use `context.AddObject("People", person)` or create a convenience method in the class to do the same.
```

Odata-docs/client/v6/basic-crud-operations.md
```diff
@@ -74,3 +78,64 @@ context.DeleteObject(person); // create a delete request
...
+
+To provide the same for our code we need to do the following.
+
+```csharp
+    // ctor from above
+    public Container(Uri serviceRoot) : base(serviceRoot)
+    {
+        this.People = base.CreateQuery<Person>("People");
+        // add the following lines
+        this.Format.LoadServiceModel= ()=> util.GetEdmModel()  /* user action that returns a valid IEdmModel instance */
```
odero:
```suggestion
        this.Format.LoadServiceModel = () => util.GetEdmModel()  /* user action that returns a valid IEdmModel instance */  
```

mumbi-o:
Closing this PR as documentation has been updated through commit [cfe8e22](https://github.com/MicrosoftDocs/OData-docs/commit/cfe8e229b141d9e50d38a4d09609f380e206c08a)

PR #116, opened 2020-03-05
Add missing examples of ODATA URLs

ArvindRSingh:
.IncludeTotalCount() // is not yet supported ref. https://github.com/OData/odata.net/issues/1334

### DIFF
--- a/Odata-docs/client/v6/query-options.md
+++ b/Odata-docs/client/v6/query-options.md
@@ -75,11 +75,15 @@ var people = context.People.OrderBy(c => c.Trips.Count);
 
 ## $skip
 
+For `GET https://host/service/EntitySet?$skip=3:
+
 ``` csharp
 var people = context.People.Skip(3);
 ```
 
 ## $top
+
+For `GET https://host/service/EntitySet?$top=3:
 
 ``` csharp
 var people = context.People.Take(3);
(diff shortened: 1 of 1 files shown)


Odata-docs/client/v6/query-options.md
```diff
@@ -75,38 +75,56 @@ var people = context.People.OrderBy(c => c.Trips.Count);
...
 ```
 
 ## A simple combined query combined
 
+For `GET https://host/service/EntitySet?$expand=Trips&$filter=FirstName eq 'Peter'&$orderby=Firstname asc&$skip=3&$top=3
+
 ``` csharp
 var people =
-    context.People.IncludeTotalCount()
+    context.People
```
xuzhg: for back-compatible, "IncludeTotalCount" is still in the codes, Right?

KenitoInc: @ArvindRSingh @xuzhg Currently we are supporting `IncludeTotalCount()`
In the next version, for backwards compatibility, we have kept `IncludeTotalCount()` and added `IncludeTotalCount(bool)`
We have also added V4 terms `IncludeCount()` and `IncludeCount(bool)`

KenitoInc: See this merged PR https://github.com/OData/odata.net/pull/1658

xuzhg:
Merged manually. Close it.

PR #138, opened 2020-07-15
Update first-odata-api.md

StephaneChevassus:
The links to download the sample code are broken.

I added (<-broken) for clarity with no means to actually display this.
Please either update the links if they still exist or perhaps we should just remove all the text to download sample code.


mumbi-o:
Thanks @StephaneChevassus for the catch. Fixed the issues you highlighted. Many thanks

PR #142, opened 2020-08-20
Typo fix Skiptoke  => SkipTake

danielabbatt:
Skiptoke  => SkipTake
I laughed when I saw this...


mumbi-o:
Thanks @danielabbatt. This has been since been updated

PR #255, opened 2022-07-27
Add docs for ODataUtf8JsonWriter

habbes:
Add articles to document the new `Utf8JsonWriter`-based `IJsonWriter` (i.e., `ODataUtf8JsonWriter`).

- Added a new article under ODataLib/Core lib: **Using Utf8JsonWriter to improve performance**
- Updated the **Dependency Injection Support** article under ODataLib/Core lib with info relevant to the new and existing JSON writer factory interfaces.


Odata-docs/odatalib/v7/using-ut8jsonwriter-for-better-performance.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Supported .NET versions
+
+This feature is only available with Microsoft.OData.Core on .NET Core 3.1 and above (i.e. .NET 6 etc.). It is not supported on Microsoft.OData.Core versions running on .NET Framework.
+
+### No support for streaming writes
+
+The default `JsonWriter` implements the `IJsonStreamWriter` and `IJsonStreamWriterAsync` interfaces. This interface allows you to pipe data directly from an input stream to the output destination. This may be useful when writing the contents of large binary file into base-64 output without buffering everything in memory first. The new `ODataUtf8JsonWriter` does not yet implement this interface, which means input from a stream will be buffered entirely in memory before being written out to the destination.
```
marabooy, line 58:
```suggestion
The default `JsonWriter` implements the `IJsonStreamWriter` and `IJsonStreamWriterAsync` interfaces. This interface allows you to pipe data directly from an input stream to the output destination. This may be useful when writing the contents of large binary file into base-64 output without buffering everything in memory first. The new `ODataUtf8JsonWriter` does not yet implement this interface, which means input from a stream will be buffered entirely in memory before being written out to the destination. If your use case requires the streaming to a stream it is recommended to use the default `JsonWriter`.
```

Odata-docs/odatalib/v7/using-ut8jsonwriter-for-better-performance.md
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.OData.Json;
+```
+
+```c#
+IContainerBuilder builder = new TestContainerBuilder();
+builder.AddDefaultODataServices();
+builder.AddService<IStreamBasedJsonWriterFactory>(sp => DefaultStreamBasedJsonWriterFactory.Default);
+```
+
+## Limitations and other considerations
```
marabooy, line 38: nit: I would Arrange the sub sections to be as follows as I believe the most pressing concern will be the .Net versions the other two I am not sure what the order will be.
```
Supported .Net versions
No Support for streaming writes
Choosing a JavaScriptEncoder
```

Odata-docs/odatalib/v7/using-ut8jsonwriter-for-better-performance.md
habbes: I've added the suggestion but rephrased it to: 

If your use case requires writing from a stream input, consider using the default `JsonWriter`.

marabooy (approved):

PR #270, opened 2022-11-14
ASP.NET Core OData 8.x - Getting Started

gathogojr:
ASP.NET Core OData 8.x - Getting Started Guide


Odata-docs/webapi-8/getting-started.md
```diff
@@ -0,0 +1,326 @@
...
+This tutorial shows how to create and run an ASP.NET Core OData 8 application
+
+You'll learn how to:  
+:white_check_mark: Create an ASP.NET Core application  
+:white_check_mark: Add Microsoft.AspNetCore.OData library package  
+:white_check_mark: Define the CLR types  
+:white_check_mark: Add an OData controller  
+:white_check_mark: Build the Edm model and configure the service  
+:white_check_mark: Run the applicaion and interact with the OData service from an API client  
+
```
asanyaga, line 23: Consider adding a prerequisites section. eg which version of Visual Studio

asanyaga, line 23: consider adding a table of the API endpoints that will be created by the tutorial. We should highlight the query options enabled by OData. See here for example https://learn.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?view=aspnetcore-6.0&tabs=visual-studio#overview-1

Odata-docs/webapi-8/getting-started.md
```diff
@@ -0,0 +1,326 @@
...
+                <EntitySet Name="Customers" EntityType="Lab01.Models.Customer" />
+            </EntityContainer>
+        </Schema>
+    </edmx:DataServices>
+</edmx:Edmx>
+```
+Notice the `Order` and `Customer` entity types as well the `Customers` entity set - nested within the `EntityContainer` element. Notice also the `Property` and `NavigationProperty` elements used to define the structure of the schema types.
+
+### Querying service data
+The query capabilities offered by OData are immense. Covering all supported capabilities is beyond the scope of this article. Let's take a look at a few:
```
asanyaga, line 210: Please add $select and also how to retrieve entity by key

Odata-docs/webapi-8/getting-started.md
```diff
@@ -0,0 +1,381 @@
...
+ms.author: jogathog
+---
+
+# Getting Started with ASP.NET Core OData 8
+**Applies To**:[!INCLUDE[appliesto-webapi](../includes/appliesto-webapi-v8.md)]
+
+This tutorial shows how to create and run an ASP.NET Core OData 8 application
+
+You'll learn how to:  
+:white_check_mark: Create an ASP.NET Core application
```
habbes, line 16: I like this format, with a list of what you'll learn and the checkmarks.

Odata-docs/webapi-8/getting-started.md
```diff
@@ -0,0 +1,381 @@
...
+:white_check_mark: Add Microsoft.AspNetCore.OData library package  
+:white_check_mark: Define the CLR types  
+:white_check_mark: Add an OData controller  
+:white_check_mark: Build the Edm model and configure the service  
+:white_check_mark: Run the OData service  
+:white_check_mark: Interact with the OData service from an API client  
+
+## Prerequisites
+- [Visual Studio 2022](https://visualstudio.microsoft.com/vs/#download) with the ASP.NET and web development workload - if target framework is [.NET Core 3.1](https://dotnet.microsoft.com/en-us/download/dotnet/3.1) or [.NET 6.0](https://dotnet.microsoft.com/en-us/download/dotnet/6.0), or
+- [Visual Studio 2019](https://visualstudio.microsoft.com/vs/older-downloads/) with the ASP.NET and web development workload - if target framework is [.NET Core 3.1](https://dotnet.microsoft.com/en-us/download/dotnet/3.1)
```
habbes, line 26: Given that .NET Core 3.1 is [going out of support on December 13th](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle), I don't think we should encourage it for new applications. If we want to document support for .NET Core 3.1, I think we should provide examples in the Guides section or another section outside the "Getting started" section.

Odata-docs/webapi-8/getting-started.md
```diff
@@ -0,0 +1,381 @@
...
+            var item = customers.SingleOrDefault(d => d.Id.Equals(key));
+
+            if (item == null)
+            {
+                return NotFound();
+            }
+
+            return Ok(item);
+        }
+    }
```
habbes, line 116: I wonder whether it would be a good idea to have a full CRUD demo in the getting started (i.e. adding methods for Create, Update and Delete). Or would that be overkill for this section?

Odata-docs/webapi-8/getting-started.md
```diff
@@ -0,0 +1,381 @@
...
+![Screenshot of configuring debugging profile in visual studio](../assets/2022-11-11-aspnet-core-odata-8-getting-started-debugging-profile.png)
+
+Press F5 to build and run the application.
+
+The following window will be displayed:
+
+![Screenshot of running aspnet core odata 8 application](../assets/2022-11-11-aspnet-core-odata-8-getting-started-run-app.png)
+
+## Interact with the OData service from an API client
+You can use an API client of your choice to interact with the OData service.
```
habbes, line 176: Should we mention the `$odata` debug endpoint or is that too much detail for a start?

Odata-docs/webapi-8/getting-started.md
```diff
@@ -0,0 +1,381 @@
...
+            app.UseEndpoints(endpoints => endpoints.MapControllers());
+        }
+    }
+}
+```
+In the _Startup.cs_ class, we define the Edm model. As part of the model, we register `Order` as an entity type and `Customers` as an entity set - `Customer` entity type also gets registered as a result. In this case, the `ODataConventionalModelBuilder` is being used to build the Edm model.
+
+We then proceed to add essential OData services by calling the `AddOData` method, in the process enabling different OData query capabilities - `$select`, `$filter`, `$orderby`, `$expand`, `count`, `$top` and `$skip`. The `AddRouteComponents` method is used to register a route passing along the Edm model to associate it with.
+
+The final step in configuring the service is adding the two middlewares responsible for matching incoming HTTP requests and dispatching those requests to the application's executable endpoints. We do this by calling `UseRouting` and `UseEndpoints` - the former adds route matching to the middleware pipeline while the latter adds endpoint execution.
```
habbes, line 161: I think it would be great to make terms like `endpoint execution` links that point to the relevant pages in the ASP.NET Core docs, cause it's not clear what "endpoint execution" means.

Odata-docs/webapi-8/getting-started.md
```diff
@@ -0,0 +1,381 @@
...
+:white_check_mark: Interact with the OData service from an API client  
+
+## Prerequisites
+- [Visual Studio 2022](https://visualstudio.microsoft.com/vs/#download) with the ASP.NET and web development workload - if target framework is [.NET Core 3.1](https://dotnet.microsoft.com/en-us/download/dotnet/3.1) or [.NET 6.0](https://dotnet.microsoft.com/en-us/download/dotnet/6.0), or
+- [Visual Studio 2019](https://visualstudio.microsoft.com/vs/older-downloads/) with the ASP.NET and web development workload - if target framework is [.NET Core 3.1](https://dotnet.microsoft.com/en-us/download/dotnet/3.1) 
+
+## Create an ASP.NET Core application
+In Visual Studio, create a new C# project from the **ASP.NET Core Empty** template. Name the project "Lab01".
+
+In the **Additional information** dialog, select either of **.NET Core 3.1 (Long Term Support)** or **.NET 6.0 (Long Term Support)** frameworks. Uncheck "Configure for HTTPS" checkbox. Click **Create**.
```
habbes, line 31: The options available for .NET Core 3.1 and .NET 6.0 are so different that I don't think it's a good idea to combine them, otherwise someone could easily make a mistake and pick options that aren't compatible with OData (e.g. minimal APIs) or with the example here.

For example, if you're on .NET 6.0 and you select the  **ASP. NET Core Empty** template, you'll be presented with the option **Do not use top-level statements**. In my case, this is unchecked by default, so you do end up with top-level statements. That means in the `Program.cs` file doesn't contain a `Program` class or `Main` method.

image

If you check that option, you get a `Program` class and `Main` method.

In both cases, there is no `Startup` class. You use a `WebApplication.CreateBuilder()` to configure the web app and services.

Someone who's new to ASP.NET Core and OData might get lost if they follow along with this example that relies on `Startup.cs` class.

I think you should stick to a specific version and template and configuration for this sample. Or be detailed about the configurations for each template and version you want to include.

asanyaga (approved):

Odata-docs/webapi-8/getting-started.md
asanyaga: We had planned a separate article for CRUD operations that's next in priority after this one

Odata-docs/webapi-8/getting-started.md
asanyaga: I think it would probably be better to include it in the CRUD article to avoid overloading the beginner with too much detail.

Odata-docs/webapi-8/getting-started.md
gathogojr: @habbes I agree with @asanyaga. If we covered `Post`, `Put`, `Patch` and `Delete` in this tutorial, it'd become too big since we'd have to provide sample requests as well. A separate article would be a better idea

Odata-docs/webapi-8/getting-started.md
gathogojr: @habbes I agree with @asanyaga about not including it in this tutorial. It'd be better to cover that in the tutorial for routing because basically what the `$odata` endpoint does is provide someone with a view of the exposed endpoints.

Odata-docs/webapi-8/getting-started.md
gathogojr: @habbes I have added the links.

Odata-docs/webapi-8/getting-started.md
gathogojr: @habbes They are actually not that different...

.NET Core 3.1 |  .NET 6.0
:------------:|:------------:
Screenshot of creating aspnetcore project targeting netcore 3.1 framework  |  Screenshot of creating aspnetcore project targeting net 6.0 framework

What I have done is add the above screenshots of creating the project using either frameworks in the doc side-by-side. I have also modified the section on how to configure the service to show how it should be done for both frameworks

habbes: @gathogojr I see the options are similar for the Empty template, but the generated code is very different. In the .NET 6.0 version, you don't have a Startup.cs file, and the default is to use top-level statements.

habbes (approved):

Odata-docs/webapi-8/getting-started.md
gathogojr: Yeah. For that reason I based the code snippet on the default option of using top-level statements

PR #272, opened 2022-12-15
Documentation for Routing in ASP.NET Core OData 8

gathogojr:
Documentation for Routing in ASP.NET Core OData 8


Odata-docs/webapi-8/entityset-routing.md
```diff
@@ -0,0 +1,523 @@
...
+For the above request to be conventionally-routed, a controller action named `Get` (or `GetShapes`) is expected, same as is expected when addressing an entity set. However, the controller action needs to be decorated with `EnableQuery` attribute. The `EnableQuery` attribute is responsible of generating the relevant query for determining the count of items:
+```csharp
+[EnableQuery]
+public ActionResult Get()
+{
+    return Ok(shapes);
+}
+```
+
+This request should return a raw value of `3`.
```
ElizabethOkerio, line 227: I think you could show the complete response.

Odata-docs/webapi-8/entityset-routing.md
```diff
@@ -0,0 +1,523 @@
...
+
+### Addressing a collection of derived entities
+The route template for this request is: `GET ~/{entityset}/{cast}`.
+
+The following request returns a collection of derived `Rectangle` entities:
+```http
+GET http://localhost:5000/odata/Shapes/EntitySetRouting.Models.Rectangle
```
ElizabethOkerio, line 234: It is worth mentioning what this `EntitySetRouting.Models` represents..

Odata-docs/webapi-8/entityset-routing.md
```diff
@@ -0,0 +1,523 @@
...
+For the above request to be conventionally-routed, a controller action named `GetFromRectangle` (or `GetShapesFromRectangle`) is expected, same as is expected when addressing a collection of derived entities. However, the controller action needs to be decorated with `EnableQuery` attribute:
+```csharp
+[EnableQuery]
+public ActionResult GetFromRectangle()
+{
+    return Ok(shapes.OfType<Rectangle>());
+}
+```
+
+This request should return a raw value of `2`.
```
ElizabethOkerio, line 285: I think showing the complete response here could provide a clear picture.

Odata-docs/webapi-8/entityset-routing.md
```diff
@@ -0,0 +1,523 @@
...
+}
+```
+
+For the above request to be conventionally-routed, a controller action named `Post` (or `PostShape`) is expected:
+```csharp
+public ActionResult Post([FromBody] Shape shape)
+{
+    shapes.Add(shape);
+
+    return Created(shape);
```
ElizabethOkerio, line 311: Is it worth mentioning why it is important to use `Created(Shape)` and not `Ok(shape)` here?

Odata-docs/webapi-8/entityset-routing.md
```diff
@@ -0,0 +1,523 @@
...
+            "Width": 4,
+            "Area": 32
+        }
+    ]
+}
+```
+
+For the above request to be conventionally-routed, a controller action named `Patch` (or `PatchShape`) is expected:
+```csharp
+public ActionResult Patch([FromBody] DeltaSet<Shape> deltaSet)
```
ElizabethOkerio, line 400: this is great advertisement of the deltaset

ElizabethOkerio: should the request body contain the @odata.context in 8.x?

Odata-docs/webapi-8/entity-routing.md
```diff
@@ -0,0 +1,508 @@
...
+| PUT | ~/{entityset}/{key} |
+| PUT | ~/{entityset}/{key}/{cast} |
+| PATCH | ~/{entityset}/{key} |
+| PATCH | ~/{entityset}/{key}/{cast} |
+| DELETE | ~/{entityset}/{key} |
+| DELETE | ~/{entityset}/{key}/{cast} |
+
+**Note:** OData routing supports canonical parentheses-style key in addition to key-as-segment. Currently, ASP.NET Core OData 8 does not support key-as-segment convention in multi-part keys scenarios.
+
+To illustrate entity routing convention, let's build a contrived OData service.
```
ElizabethOkerio, line 35: is there a simpler word that can be used in place of contrived in all the pages?

Odata-docs/webapi-8/action-routing.md
```diff
@@ -0,0 +1,423 @@
...
+OData action routing convention supports the following route templates:
+
+| Request Method | Route Template |
+|----------------|----------------|
+| POST | ~/{entityset}\|{singleton}/{action} |
+| POST | ~/{entityset}\|{singleton}/{cast}/{action} |
+| POST | ~/{entityset}/{key}/{action} |
+| POST | ~/{entityset}/{key}/{cast}/{action} |
+
+**Note:** OData routing supports canonical parentheses-style key in addition to key-as-segment. Currently, ASP.NET Core OData 8 does not support key-as-segment convention in multi-part keys scenarios.
```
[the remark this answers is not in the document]
ElizabethOkerio: How does key-as-segment in multi-part keys scenario represented? Is it supported in 7.x , will it be supported in 8.x or not?

Odata-docs/webapi-8/entityset-routing.md
gathogojr: @ElizabethOkerio Please note that the `DeltaSet<T>` in 8.0 only serves to allow someone to send a collection of `Delta<T>`. Support for bulk operations is not implemented. It therefore doesn't expect the `@odata.context` property in the payload

Odata-docs/webapi-8/entityset-routing.md
gathogojr: @ElizabethOkerio Actually, the raw value 3 is all there is in the response. I could however render the response in a code block so it's easier for the reader to relate with. As follows:

```console
3
```

Odata-docs/webapi-8/routing-overview.md
```diff
@@ -0,0 +1,346 @@
...
+
+    [HttpGet("odata/Customers({key})")]
+    public ActionResult Details([FromRoute] int key)
+    {
+        return Ok(new Customer { Id = key, Name = $"Customer {key}" });
+    }
+}
+```
+
+In this case, the controller action is explicitly matched to the hard-coded route template. The route prefix is included to associate the route template with the configured route.
```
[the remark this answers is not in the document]
gathogojr: It'll be a non-OData endpoint since it won't be associated with any OData route and Edm model.

Odata-docs/webapi-8/routing-overview.md
```diff
@@ -0,0 +1,346 @@
...
+        {
+            "name": "Customers",
+            "kind": "EntitySet",
+            "url": "Customers"
+        }
+    ]
+}
+```
+
+Resources exposed by an OData service are addressable by the corresponding **resource path**. Examples of resources might customers, a single customer, and so forth. Examples of addressable aspects of these resources as exposed by the data model might be: collections of entities, a single entity, properties, links, operations, and so on. As an illustration, to address a collection of customer entities via the entity set, the relevant resource path would be `Customers`.
```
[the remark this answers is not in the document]
gathogojr: "addressable aspects" = "things about the resources" that you can address or interact with

Odata-docs/webapi-8/entityset-routing.md
```diff
@@ -0,0 +1,545 @@
...
+public class Circle : Shape
+{
+    public double Radius { get; set; }
+}
+```
+
+## Edm model and service configuration
+The logic for building the Edm model and configuring the OData service is as follows:
+
+```csharp
```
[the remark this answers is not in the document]
gathogojr: Yeah. It was a test for rendering tabs. I fixed it

Odata-docs/webapi-8/routing-overview.md
```diff
@@ -0,0 +1,346 @@
...
+## OData URL components
+An OData URL has at most 3 significant parts: the _service root_, the _resource path_, and the _query options_.
+```
+http://localhost:5000/odata/Customers/?$select=Name
+\_________________________/\________/\____________/
+            |                   |           |
+      service root       resource path  query options
+```
+
+The **service root** is a combination of the root URL and the route prefix where one is configured. The service root identifies the root of an OData service. A GET request to this URL returns the format-specific service document.
```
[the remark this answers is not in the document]
habbes: I think that works

Odata-docs/webapi-8/entity-routing.md
gathogojr: @ElizabethOkerio I can use sample :)

Odata-docs/webapi-8/entityset-routing.md
```diff
@@ -0,0 +1,545 @@
...
+        new Circle { Id = 2, Radius = 3.5, Area = 38.5 },
+        new Rectangle { Id = 3, Length = 8, Width = 5, Area = 40 }
+    };
+}
+```
+
+## Routing conventions for entity sets
+In this section we cover the conventions for entity set routing and the controller actions (endpoints) required for the requests to be routed successfully.
+
+### Addressing an entity set
```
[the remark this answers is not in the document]
gathogojr: @habbes Fixed

Odata-docs/webapi-8/entityset-routing.md
```diff
@@ -0,0 +1,545 @@
...
+public ActionResult Get()
+{
+    return Ok(shapes);
+}
+```
+
+This request should return a raw value of `3`.
+
+### Addressing a collection of derived entities
+The route template for this request is: `GET ~/{entityset}/{cast}`.
```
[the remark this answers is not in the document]
gathogojr: @habbes I have mentioned that right below the table that lists the supported route templates

Odata-docs/webapi-8/entity-routing.md
```diff
@@ -0,0 +1,471 @@
...
+| GET | ~/{entityset}/{key} |
+| GET | ~/{entityset}/{key}/{cast} |
+| PUT | ~/{entityset}/{key} |
+| PUT | ~/{entityset}/{key}/{cast} |
+| PATCH | ~/{entityset}/{key} |
+| PATCH | ~/{entityset}/{key}/{cast} |
+| DELETE | ~/{entityset}/{key} |
+| DELETE | ~/{entityset}/{key}/{cast} |
+
+**Note:** OData routing supports canonical parentheses-style key in addition to key-as-segment. Currently, ASP.NET Core OData 8 does not support key-as-segment convention in multi-part keys scenarios.
```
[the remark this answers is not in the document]
gathogojr: @habbes I have modified the statement to provide examples for both canonical parentheses-style key and key-as-segment

Odata-docs/webapi-8/entity-routing.md
```diff
@@ -0,0 +1,471 @@
...
+    "Id": 1,
+    "Area": 28.0,
+    "Length": 7.0,
+    "Width": 4.0
+}
+```
+
+Since `Rectangle` is a derived type the object in the response contain an `@odata.type` property specifying the type of the entity.
+
+### Addressing a single derived entity
```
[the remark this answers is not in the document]
gathogojr: @gathogojr Fixed using "Retrieving"

Odata-docs/webapi-8/entity-routing.md
```diff
@@ -0,0 +1,471 @@
...
+The route templates for this request are:
+- `GET ~/{entityset}({key})/{cast}`.
+- `GET ~/{entityset}/{key}/{cast}`
+
+The following request returns a single `Circle` derived entity with the key value of 2:
+```http
+GET http://localhost:5000/odata/Shapes(2)/EntityRouting.Models.Circle
```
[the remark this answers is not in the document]
gathogojr: @habbes @habbes I have mentioned that right below the table that lists the supported route templates

Odata-docs/webapi-8/entity-routing.md
```diff
@@ -0,0 +1,471 @@
...
+Here's the request body:
+```json
+{
+    "@odata.type": "#EntityRouting.Models.Rectangle",
+    "Width": 6,
+    "Area": 48
+}
+```
+
+For the above request to be conventionally-routed, a controller action named `Patch` (or `PatchShape`) is expected. The controller action should accept two parameters - the first is the key of the target entity and the second is a parameter of type `Delta<Shape>` decorated with `FromBody` attribute:
```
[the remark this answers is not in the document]
gathogojr: @habbes I have added links to the docs for `Delta<T>` (or `DeltaSet<T>` as the case may be)

Odata-docs/webapi-8/action-routing.md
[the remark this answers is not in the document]
gathogojr: @habbes Fixed

Odata-docs/webapi-8/action-routing.md
```diff
@@ -0,0 +1,423 @@
...
+```csharp
+var computeSalaryFunction = modelBuilder.Action("ComputeSalary");
+computeSalaryFunction.Parameter<decimal>("hourlyRate");
+computeSalaryFunction.Parameter<int>("hoursWorked");
+computeSalaryFunction.Returns<decimal>();
+```
+
+An unbound action can be placed in any controller in the application. To avoid confusion, you can create a controller unassociated with any entity set to serve as a home for your unbound operations.
+
+To associate an unbound action with a configured non-default/non-empty route, the route should be placed on the controller action:
```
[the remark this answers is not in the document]
gathogojr: @habbes Modified the paragraph to explain what default route is

Odata-docs/webapi-8/entityset-routing.md
gathogojr: @ElizabethOkerio Fixed

Odata-docs/webapi-8/entityset-routing.md
gathogojr: @ElizabethOkerio Fixed this by specifying in the **Models** section that the classes are defined under the `EntitySetRouting.Models` namespace. Did similar thing across other files

Odata-docs/webapi-8/entityset-routing.md
gathogojr: @ElizabethOkerio I have explained that the `Created` method generates the location of the created entity and returns it as a response header

Odata-docs/webapi-8/entityset-routing.md
```diff
@@ -0,0 +1,536 @@
...
+The route template for this request is: `GET ~/{entityset}/$count`.
+
+To address the raw value of the number of items in an entity set, append `/$count` to that entity set's URL.
+
+The following request returns a count of items in the `Shapes` entity set:
+```http
+GET http://localhost:5000/odata/Shapes/$count
+```
+
+For the above request to be conventionally-routed, a controller action named `Get` (or `GetShapes`) is expected, same as is expected when addressing an entity set. However, the controller action needs to be decorated with `EnableQuery` attribute. The `EnableQuery` attribute is responsible of generating the relevant query for determining the count of items:
```
ElizabethOkerio, line 223: If you had enabled the query options when setting up the model do you still have to decorate the controller action methods with `EnableQuery`

Odata-docs/webapi-8/function-routing.md
```diff
@@ -0,0 +1,365 @@
...
+## Prerequisites
+
+[!INCLUDE[](../includes/appliesto-webapi-v8-net-prereqs-vs.md)]
+
+## Packages
+
+[!INCLUDE[](../includes/appliesto-webapi-v8-pkg-install.md)]
+
+## Models
+The following are the models for the OData service. Define them under `FunctionRouting.Models` namespace:
```
ElizabethOkerio, line 42: does this mean you set the `FunctionRouting.Models` namespace? I think it is confusing.. The statement could be left out.

Odata-docs/webapi-8/function-routing.md
```diff
@@ -0,0 +1,365 @@
...
+
+### Invoking a function bound to an entity set or singleton
+The route template for this request is: `GET ~/{entityset}|{singleton}/{function}`
+
+The following request invokes the `GetHighestRating` function bound to the `Employees` entity set. The URL for the function is the function name appended to the entity set's URL:
+```http
+GET http://localhost:5000/odata/Employees/GetHighestRating()
+```
+
+An alternative URL for invoking the same function is: `http://localhost:5000/odata/Employees/Default.GetHighestRating()`.
```
ElizabethOkerio, line 194: If we defined the model elements in the `FunctionRouting.Models` namespace.. why do we use `Default` here?

Odata-docs/webapi-8/function-routing.md
```diff
@@ -0,0 +1,365 @@
...
+- `POST ~/{entityset}/{key}/{cast}/{function}`
+
+The following request invokes the `GetBonus` function bound to employee 5 (a manager). The URL for the function is the function name appended to the derived entity's URL:
+```http
+GET http://localhost:5000/odata/Employees(5)/FunctionRouting.Models.Manager/GetBonus()
+```
+
+An alternative URL for invoking the same function is: `http://localhost:5000/odata/Employees(5)/FunctionRouting.Models.Manager/Default.GetBonus()`.
+
+For the above request to be conventionally-routed, a controller action named `GetBonusOnManager` that accepts no parameter is expected. The controller action should be decorated with a `HttpGet` attribute.
```
ElizabethOkerio, line 300: here you've written it accepts no parameter but has a key parameter?

Odata-docs/webapi-8/entityset-routing.md
gathogojr: @ElizabethOkerio In _Startup.cs_ we only enable the features. It's the `EnableQuery` attribute that does the magic. It is required. Alternatively, someone can add the `ODataQueryOptions<T>` parameter to the method and then do `options.ApplyTo(IQueryable)` in the method body. This alternative way should be covered in the OData Query Options doc. I didn't want to introduce it in the context of routing to avoid confusion

Odata-docs/webapi-8/function-routing.md
gathogojr: This is the namespace for the C# classes. It's mentioned under the **Models** section so it's relevant to the data models. The section for building Edm model is different and we don't change the Edm model namespace so it'll assume the value of `Default`.

Odata-docs/webapi-8/function-routing.md
gathogojr: @ElizabethOkerio In the section for building Edm model we don't change the Edm model namespace so it'll assume the value of `Default`. The structured elements always assume the namespace that the classes are defined in. Here's the $metadata for the actual service:
```xml
<?xml version="1.0" encoding="utf-8"?>
<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
    <edmx:DataServices>
        <Schema Namespace="FunctionRouting.Models" xmlns="http://docs.oasis-open.org/odata/ns/edm">
            <EntityType Name="Employee">
                <Key>
                    <PropertyRef Name="Id" />
                </Key>
                <Property Name="Id" Type="Edm.Int32" Nullable="false" />
                <Property Name="Name" Type="Edm.String" />
                <Property Name="PerfRating" Type="Edm.Int32" Nullable="false" />
            </EntityType>
            <EntityType Name="Manager" BaseType="FunctionRouting.Models.Employee">
                <Property Name="Bonus" Type="Edm.Decimal" Nullable="false" Scale="Variable" />
            </EntityType>
        </Schema>
        <Schema Namespace="Default" xmlns="http://docs.oasis-open.org/odata/ns/edm">
            <Function Name="GetHighestRating" IsBound="true">
                <Parameter Name="bindingParameter" Type="Collection(FunctionRouting.Models.Employee)" />
                <ReturnType Type="Edm.Int32" Nullable="false" />
            </Function>
            <Function Name="GetRating" IsBound="true">
                <Parameter Name="bindingParameter" Type="FunctionRouting.Models.Employee" />
                <ReturnType Type="Edm.Int32" Nullable="false" />
            </Function>
            <Function Name="GetHighestBonus" IsBound="true">
                <Parameter Name="bindingParameter" Type="Collection(FunctionRouting.Models.Manager)" />
                <ReturnType Type="Edm.Decimal" Nullable="false" Scale="Variable" />
            </Function>
            <Function Name="GetBonus" IsBound="true">
                <Parameter Name="bindingParameter" Type="FunctionRouting.Models.Manager" />
                <ReturnType Type="Edm.Decimal" Nullable="false" Scale="Variable" />
            </Function>
            <Function Name="GetSalary">
                <Parameter Name="hourlyRate" Type="Edm.Decimal" Nullable="false" Scale="Variable" />
                <Parameter Name="hoursWorked" Type="Edm.Int32" Nullable="false" />
                <ReturnType Type="Edm.Decimal" Nullable="false" Scale="Variable" />
            </Function>
            <EntityContainer Name="Container">
                <EntitySet Name="Employees" EntityType="FunctionRouting.Models.Employee" />
                <FunctionImport Name="GetSalary" Function="Default.GetSalary" IncludeInServiceDocument="true" />
            </EntityContainer>
        </Schema>
    </edmx:DataServices>
</edmx:Edmx>
```

There are two schema elements. The first one has the namespace `FunctionRouting.Models` while the second one has the namespace `Default` which is the, well, default. The actions and functions are under the `Default` namespace.

Odata-docs/webapi-8/routing-overview.md
```diff
@@ -0,0 +1,365 @@
...
+- [`ActionRoutingConvention`](/odata/webapi-8/action-routing) - routes requests for Edm actions
+- `OperationImportRoutingConvention` - routes requests for operation imports
+- `PropertyRoutingConvention` - routes requests for structural properties
+- [`NavigationRoutingConvention`](/odata/webapi-8/navigation-routing) - routes requests for navigation properties
+- [`RefRoutingConvention`](/odata/webapi-8/ref-routing) - routes requests for entity reference links
+- `AttributeRoutingConvention` - represents a routing convention for attribute-routed requests
+
+A in-depth walkthrough for the core routing conventions can be found in the relevant tutorial.
+
+At application startup, OData path templates for all potential endpoints are built. The OData routing conventions above are "executed" to generate a route table that maps each controller action (endpoint) to an OData route template.
```
ElizabethOkerio, line 87: all conventions get executed at startup?

Odata-docs/webapi-8/function-routing.md
gathogojr: @ElizabethOkerio Would it help if I removed that statement and then replaced this:
```csharp
public class Employee
{
    public int Id { get; set; }
    public string Name { get; set; }
    public int PerfRating { get; set; }
}
```
with this:
```
namespace FunctionRouting.Models
{
    public class Employee
    {
        public int Id { get; set; }
        public string Name { get; set; }
        public int PerfRating { get; set; }
    }
}
```
?
I had added that statement because there's a review comment where you had said I should mention about the namespace. This second way may be less confusing but someone will still have to understand that this namespace finds it's way to the Edm model.
I think for now I'll implement my suggestion here since it needs less explanation

Odata-docs/webapi-8/routing-overview.md
```diff
@@ -0,0 +1,365 @@
...
+```csharp
+public interface IODataControllerActionConvention
+{
+    int Order { get; }
+    bool AppliesToController(ODataControllerActionContext context);
+    bool AppliesToAction(ODataControllerActionContext context);
+}
+```
+
+- `Order` is used to specify the execution order of the convention. Conventions are executed in ascending numeric value of the `Order` property.
```
ElizabethOkerio, line 99: is it worth the mentioning the order of the various conventions here?

Odata-docs/webapi-8/routing-overview.md
```diff
@@ -0,0 +1,365 @@
...
+        {
+            "name": "Customers",
+            "kind": "EntitySet",
+            "url": "Customers"
+        }
+    ]
+}
+```
+
+Resources exposed by an OData service are addressable by the corresponding **resource path**. Examples of resources might customers, a single customer, and so forth. Examples of addressable aspects of these resources as exposed by the data model might be: collections of entities, a single entity, properties, links, operations, and so on. As an illustration, to address a collection of customer entities via the entity set, the relevant resource path would be `Customers`.
```
ElizabethOkerio, line 246:
```suggestion
Resources exposed by an OData service are addressable by the corresponding **resource path**. Examples of resources might be or Examples of resource are: customers, a single customer, and so forth. Examples of addressable aspects of these resources as exposed by the data model might be: collections of entities, a single entity, properties, links, operations, and so on. As an illustration, to address a collection of customer entities via the entity set, the relevant resource path would be `Customers`.
```

Odata-docs/webapi-8/function-routing.md
gathogojr: @ElizabethOkerio Good catch! Thanks

Odata-docs/webapi-8/routing-overview.md
```diff
@@ -0,0 +1,365 @@
...
+        }
+    ]
+}
+```
+
+The **query options** part of an OData URL specifies three types of information: system query options, customer query options and parameter aliases.
+
+System query options are query string parameters that control the amount and order of the data returned for the resource identified by the URL. The names of all system query options are prefixed with a dollar (`$`) symbol by default. However, the OData standard does not make it mandatory to prefix system query options with a dollar symbol so ASP.NET Core OData 8 allows you to relax or enforce the requirement for the prefix. Examples of system query options are: `$filter`, `$orderby`, and `$top`.
+
+Custom query options provide an extensible mechanism for service-specific information to be placed in a URL query string. Customer query options MUST NOT begin with a `$` or `@` character.
```
ElizabethOkerio, line 272: what are custom query options?

ElizabethOkerio, line 272:
```suggestion
Custom query options provide an extensible mechanism for service-specific information to be placed in a URL query string. Custom query options MUST NOT begin with a `$` or `@` character.
```

Odata-docs/webapi-8/routing-overview.md
```diff
@@ -0,0 +1,365 @@
...
+        },
+        {
+            "Id": 2,
+            "Name": "Customer 2"
+        }
+    ]
+}
+```
+
+The **query options** part of an OData URL specifies three types of information: system query options, customer query options and parameter aliases.
```
ElizabethOkerio, line 268:
```suggestion
The **query options** part of an OData URL specifies three types of information: system query options, custom query options and parameter aliases.
```

ElizabethOkerio: Is it custom or customer?

Odata-docs/webapi-8/routing-overview.md
ElizabethOkerio, line 272:
```suggestion
Custom query options provide an extensible mechanism for service-specific information to be placed in a URL query string. Custom query options MUST NOT begin with a `$` or `@` character.
```

Odata-docs/webapi-8/routing-overview.md
```diff
@@ -0,0 +1,365 @@
...
+        // ...
+    }
+}
+```
+
+#### `ODataAttributeRouting` on controller action
+```csharp
+public class CustomersController : ControllerBase
+{
+    [ODataAttributeRouting]
```
ElizabethOkerio, line 314: I thought in 8.x we can use aspnetcore's Route attribute and we don't need to use the ODataAttributeRouting attribute.

Odata-docs/webapi-8/routing-overview.md
ElizabethOkerio: I noticed you didn't mention some of the conventions: MetadataRoutingConvention, OperationImportRoutingConvention and PropertyRoutingConvention

gathogojr: @ElizabethOkerio "executed", "run", "triggered", etc. Basically what we do is try to apply each convention to the controller and controller actions defined in the project with the output of that exercise being the route table that we render when someone visits the `/$odata` page. This is unlike 7.x where we used to try and find a match upon receiving the request. In 8.x it's just a matter of looking up from the pre-generated route table. I don't know if there's a better word that I can use in place of "executed". It's actually the reason I placed it between quotes

Odata-docs/webapi-8/entityset-routing.md
ElizabethOkerio: What are the features? Also you can do something like .Count().Expand.Select(). This is just for my own knowledge.

Odata-docs/webapi-8/routing-overview.md
gathogojr: I'll add an example

gathogojr: Done

Odata-docs/webapi-8/routing-overview.md
gathogojr: Already, talking about `AppliesToAction`, `AppliesToController` and `Order` felt like it was making the doc too complex. But I needed to provide a big of info about what happens in the background. I think list the order of various conventions will only add to the complexity.

Odata-docs/webapi-8/routing-overview.md
gathogojr: @ElizabethOkerio I'll do a separate doc for `PropertyRoutingConvention`. There are far too many route templates covered by that convention and I realized the rest would get seriously delayed. There other two are trivial: `MetadataRoutingConvention` is used to route the `/$metadata` requests. But I'll make sure I cover them in some sort of way

ElizabethOkerio (approved):

Odata-docs/webapi-8/entityset-routing.md
gathogojr: @ElizabethOkerio Yeah... `.Count().Expand.Select()` are the OData query features I'm talking about here. We actually support a method `EnableQueryFeatures(int?)` that enables them all instead of doing `.Select().Filter().OrderBy().Expand().Count().SkipToken().SetMaxTop(null)`. I have used that method in all the doc pages in this PR to keep the statement short

habbes (approved):

Odata-docs/webapi-8/fundamentals/routing-overview.md
```diff
@@ -0,0 +1,365 @@
...
+---
+
+# Routing in ASP.NET Core OData 8
+**Applies To**:[!INCLUDE[appliesto-webapi](../../includes/appliesto-webapi-v8.md)]
+
+This tutorial covers routing in ASP.NET Core OData 8. This tutorial assumes a basic understanding of how to create and run an ASP.NET Core OData 8 application. If you're unfamiliar with ASP.NET Core OData 8, you may want to go through the [Getting started](/odata/webapi-8/getting-started) tutorial.
+
+## Prerequisites
+
+[!INCLUDE[](../../includes/appliesto-webapi-v8-net-prereqs-vs.md)]
```
asanyaga, line 17: @gathogojr  do you think that we might want to point to the routing article for ASP.NET? I found it hard to follow along the OData concepts until after I read this https://learn.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-6.0

Odata-docs/webapi-8/fundamentals/entityset-routing.md
```diff
@@ -0,0 +1,545 @@
...
+The route template for this request is: `GET ~/{entityset}/$count`.
+
+To address the raw value of the number of items in an entity set, append `/$count` to that entity set's URL.
+
+The following request returns a count of items in the `Shapes` entity set:
+```http
+GET http://localhost:5000/odata/Shapes/$count
+```
+
+For the above request to be conventionally-routed, a controller action named `Get` (or `GetShapes`) is expected, same as is expected when [retrieving an entity set](#retrieving-an-entity-set). However, the controller action needs to be decorated with `EnableQuery` attribute. The `EnableQuery` attribute is responsible for generating the relevant query for determining the number of items:
```
asanyaga, line 232: @gathogojr how does the EnableQuery attribute interact with the EnableQueryFeatures configuration?

PR #280, opened 2023-01-09
Web API 8 fundamentals overview

habbes:


Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+            }
+        }
+    ]
+}
+
+```
+
+**Response**
+
+The response body contains an array of responses, on for each request. Each response includes the status, headers, body and other metadata that allows the client to process the response.
```
gathogojr, line 509:
```suggestion
The response body contains an array of responses, one for each request. Each response includes the HTTP status code, headers, body and other metadata that allows the client to process the response.
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+                "Color": "Blue",
+                "CreatedDate": "2021-12-27T09:12:08-08:00",
+                "UpdatedDate": null
+            }
+        }
+    ]
+}
+```
+
+This batch request contained the requests:
```
gathogojr, line 569:
```suggestion
This batch request contained the following requests:
```

gathogojr, line 569: I however don't know if it's important to list the requests here yet they can be inferred from the batch request body

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+- `GET http://localhost:5000/Products/1`
+- `GET http://localhost:5000/People?$select=FirstName&$top=2`
+- `PATCH http://localhost:5000/Products/2`
+
+The batch handler breaks down the batch request into separate requests, executes them and combines them in a single batch response. Each request in the batch is routed and handled as if it had been made directly in the client. This means that each request will automatically be routed to the matching controller method. If there are any dependencies between requests (e.g. one request relies on the response of another), the batch handler will ensure they are executed in the right order.
```
gathogojr, line 574:
```suggestion
The batch handler breaks down the batch request into separate requests, executes them and combines them in a single batch response. Each request in the batch is routed and handled as if it had been made directly from the client. This means that each request will automatically be routed to the relevant controller action. If there are any dependencies between requests (e.g. one request relies on the response of another), the batch handler will ensure they are executed in the right order.
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+
+For more information about JSON batching, [visit this article](/odata/odatalib/json-batch).
+For examples using the `multipart/mixed` format, [visit this article](/odata/webapi/batch).
+<!-- TODO: replace these links with a link to an upto-date article dedicate to OData batching in ASP.NET Core OData 8 once that article has been written (see work item 1980957)-->
```
gathogojr, line 578:
```suggestion
<!-- TODO: replace these links with a link to an upto-date article dedicated to OData batching in ASP.NET Core OData 8 once that article has been written (see work item 1980957)-->
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+}
+```
+
+---
+
+We register OData services using the `services.AddControllers().AddOData()` extension method. The `AddOData()` method accepts a callback action that allows us to configure different aspects of the OData service using the options argument (which is an instance of [`ODataOptions`](/dotnet/api/microsoft.aspnetcore.odata.odataoptions)).
+
+In this sample we're calling the `options.AddRouteComponents()` method which is used to register and configure and the components and services that make up our OData API. We pass an EDM model that describes the service schema to the `options.AddRouteComponents()`.
+
+The EDM model is created with the lines:
```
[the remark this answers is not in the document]
habbes: Thanks. FYI, this is one of the places I thought snippet highlighting would come in handy.

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+This will create OData services with the following url:
+- `<baseUrl>/api1` using the schema defined by `edmModel1`.
+- `<baseUrl>/api2` using the schema defined by `edmModel2`.
+
+## Routing
+
+> [!IMPORTANT]
+> ASP.NET Core OData routing is based on controllers and therefore does not support [minimal APIs](/aspnet/core/fundamentals/minimal-apis/overview).
+
+The OData specification defines a number of conventions for routing and endpoints for accessing various resources based on REST best-practices. For example, if you have an entity set called `Customers` based on the `Customer` entity type, the service will be expected to expose a `GET /Customers` endpoint that returns a collection of `Customer` entities, it will also be expected to expose a `GET /Customers({id})` endpoint that returns a customer based on ID.
```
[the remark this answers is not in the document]
habbes: I used the term "expected" to denote that it's not mandatory, but based on convention (and the standard), that's how you'd do it. You _can_ use a function, but it's not what's expected. Let me think of a better way to phrase it.

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+    }
+
+    public ActionResult<IQueryable<Customer>> Get()
+    {
+        return Ok(_db.Customers);
+    }
+
+    public ActionResult<Customer> Get(int id)
+    {
+        return _db.Customers.Find(id);
```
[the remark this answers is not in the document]
habbes: Nice catch.

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+    public ActionResult<Customer> Post(Customer customer)
+    {
+        _db.Customers.Add(customer);
+        _db.SaveChanges();
+        return Created(customer);
+    }
+}
+```
+
+- The `Get()` controller action will be mapped to the `GET /Customers` endpoint.
```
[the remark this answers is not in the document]
habbes: I guess in general API jargon, endpoints refer to the URLs/paths that clients can hit to request a resource (e.g. [see this discussion](https://stackoverflow.com/questions/18929325/what-is-meant-by-an-api-endpoint)). I think the term is overloaded, and ASP.NET Core may have other uses of the term. Since I have not used the term `route template` before in this article, maybe I can use `route` instead if `endpoint` is ambiguous.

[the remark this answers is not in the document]
habbes: The endpoint mapping table doesn't specify what is an endpoint (I mean the term "endpoint" doesn't appear in any of the table headings).

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+- The `Get(int id)` controller action will be mapped to the `GET /Customers({id})` and `GET /Customers/{id}` endpoints
+- The `Post(Customer customer)` controller action will be mapped to the `POST /Customers` endpoint.
+
+It is up to the developer to implement the actual logic to perform the expected operation and return the expected response. For example, it is up to the developer to ensure that the `Get()` method returns a collection of `Customer` objects, the library does not enforce this. This means that it's possible for an application written with ASP.NET Core OData not to be compliant with OData specifications or convnetions.
+
+The library supports routing for common OData endpoints like entity sets, singletons, actions, functions, nested properties, etc. You can specify an endpoint for a specific controller action explicitly using ASP.NET Core `[Route]` attributes. You can also define your own routing conventions to extend the built-in conventions or define custom routing logic.
```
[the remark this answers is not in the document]
habbes: Good question. I've removed the "nested".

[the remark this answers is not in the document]
habbes: You're right. Let me go with "attribute routing"

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+
+Visit the [routing documentation](/odata/webapi-8/fundamentals/routing-overview) to learn more about routing.
+
+## Query handling
+
+OData defines various query options that clients can add to an OData request URL to query and transform the response OData. Here are some of the common query options:
```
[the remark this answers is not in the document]
habbes: I think I meant "OData response". I considered `$filter`, `$top`, `$skip`, `$orderby` as query and the `$select`, `$expand`, `$apply` as transforms. By query I mean operations that allow you to narrow down the data to exactly what you want (which includes limiting the result set).

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+- `$filter` allows you to filter results based on specific conditions: `/Products?$filter=Price lt 1000 and Category eq 'Electronics'` will only return electronic products whose price is less than 1,000.
+- `$select` allows you to limit the fields that are included in the response: `/Products?$select=Name,Category,DateCreated` will only include the `Name`, `Category`, `DateCreated` fields in the response
+- `$expand` allows you to join and include related data in the response: `/OrderItems?$expand=Product` will return order items and include the related product in the response
+- `$skip` and `$top` allow you to limit the number of records in the response and can be used for client-driven pagination: `/Products$top=10&$skip=20&$orderby=Name` returns the third page of 10 products sorted by name.
```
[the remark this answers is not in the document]
habbes: I added `$orderby` as a separate list item.

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+You can enable all supported query options in one go using the `EnableQueryFeatures` method:
+
+```c#
+services.AddControllers().AddOData(
+    options => options.EnableQueryFeatures(100).AddRouteComponents(edmModel));
+```
+
+The argument passed to `EnableQueryFeatures` is the max top value. You can set it to `null` if you don't want to impose a limit on the value passed to `$top`.
+
+After enabling query options on the service, you also have to enable it on specific endpoints where you want to support query options. You do this by adding the [`[EnableQuery]`](/dotnet/api/microsoft.aspnetcore.odata.query.enablequeryattribute) attribute on the corresponding controller actions. Without specifying `[EnableQuery]` explicitly, OData query options will be ignored for that endpoint.
```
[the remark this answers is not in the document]
habbes: I've rephrased it. Also rephrased it to make it clearer that `[EnableQuery]` is one option. I haven't mentioned using `ODataQueryOptions` directly since that can be covered in the query handling docs.

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+ASP.NET Core OData uses JSON format for its requests and responses (the `$metadata` endpoint returns the schema in XML format by default). The OData JSON format may include metadata in the payload such as type information that makes it possible for clients and server to know how to handle the payload. This metadata is usually expressed in the form of "annotations" which are extra fields in the payload that start with `@odata.` prefix.
+
+For example. Let's say we have an OData service that has a `Customers` entity set which returns entities of type `Customer`.
+
+A request like `GET /Customers`, we would a get a response like
+
+```json
+{
+    "@odata.context": "http://localhost:5000/$metadata#Customers",
+    "value": [
```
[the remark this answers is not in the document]
habbes: I didn't realize it was in descending order. Maybe I inadvertently copied the response from a request that had an `$orderby`.

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+
+To achieve this, ASP.NET Core OData uses customer input and output formatters to handle serialization and deserialization. These formatters know how to handle the OData-specific JSON format. It may also perform some validation to ensure the payloads match OData's specification. For example, it may validate that the properties in the request and response are actually defined in the OData model.
+
+## Batch requests
+
+OData allows you to group multiple operations in a single HTTP request using a batch request. The response to this batch request contains the responses to the individual operations that were in the batch request. Batch requests allow you to reduce the number of round trips between client and server and can improve the performance and scalability of a service.
```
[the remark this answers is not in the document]
habbes: What about "in a single HTTP request, called a batch request" to avoid assuming that the reader knows what a batch request is.

Odata-docs/webapi-8/fundamentals/overview.md
habbes: I opted to remove this list altogether since it's redundant.

habbes: You're right. It's redundant.

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+    </edmx:DataServices>
+</edmx:Edmx>
+```
+
+
+Clients can consume this metadata to know what types and endpoints the service exposes. Many OData clients use this metadata to generate strongly-typed client code. Tools like the [OData Connected Service](/odata/connectedservice/getting-started) and [OData CLI](/odata/odatacli/getting-started) can generate client code based on an OData schema.
+
+## Registering OData services
+
+It is possible to service multiple OData APIs in the same ASP.NET Core application. You register an OData API using the `options.AddRouteComponents` method. This method registers the EDM model, base route and components for a specific OData service.
```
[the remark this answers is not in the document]
habbes: Removed the redundant statement here.

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,578 @@
...
+
+In the example above, we register a single OData service. This is hosted on the "default" route, i.e. on the root path `/`. This means that the paths on this OData service look like `/Customers`, `/Products`, `/$metadata`.
```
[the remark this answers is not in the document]
habbes: I meant the example in the intro of the document (at the beginning), which doesn't have a section header.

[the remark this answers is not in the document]
habbes: I have rephrased the statement as "The sample code in the introduction..."

Odata-docs/webapi-8/fundamentals/overview.md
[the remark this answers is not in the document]
habbes: I've rephrased it as "the convention is to..."

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,581 @@
...
+
+ms.date: 1/5/2023
+---
+# Fundamentals Overview
+**Applies To**:[!INCLUDE[appliesto-webapi](../../includes/appliesto-webapi-v8.md)]
+
+This article provides an overview of the fundamental concepts for building ASP.NET Core OData apps.
+
+ASP.NET Core OData is a .NET library that sits on top of ASP.NET Core to help you build REST APIs. The ASP.NET Core OData library consists of a set of services and middleware that hook into ASP.NET Core's request pipeline to provide features like routing, query handling and serialization based on OData specifications and conventions. All these concepts will be covered in depth in separate articles, this just provides an overview.
+
```
asanyaga, line 16: @habbes consider including a list that outlines the concepts that are detailed in the article e.g. 1.create edm model. 2. register services 3.configure querying etc...
i think it would be useful to a complete beginner to OData and even ASP.NET have a logical view of how to set up the various bits for an OData service

habbes: Do you mean like a list in the introduction, or something like a "table of contents".

If it's the latter, then it's auto-generated:

image

asanyaga: something like this 
image

habbes: I have gone through a few samples in the ASP.NET Core docs and it appears this this outline is mostly used in tutorials-like article where the entire article takes you through a sequence of steps to achieve a goal, e.g.: 
- https://learn.microsoft.com/en-us/aspnet/core/tutorials/razor-pages/razor-pages-start?view=aspnetcore-6.0&tabs=visual-studio
- https://learn.microsoft.com/en-us/aspnet/core/tutorials/first-web-api?view=aspnetcore-6.0&tabs=visual-studio
- https://learn.microsoft.com/en-us/aspnet/core/getting-started/?view=aspnetcore-6.0&tabs=windows

But it's not include in articles that provide a guide on conceptual topics:
- https://learn.microsoft.com/en-us/aspnet/core/fundamentals/?view=aspnetcore-6.0&tabs=windows
- https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/?view=aspnetcore-6.0&tabs=windows

This is not a solid argument for not including that part. And I'm not against the suggestion. I'm just not sure what to write, what the exact outline should be. What I have in mind is to list the headings, but that would be a duplicate of the outline already auto-generated in the sidebar. If you think this is fine, I can go ahead and add it.

habbes: cc @asanyaga

asanyaga (approved):

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,581 @@
...
+```dotnetcli
+dotnet add package Microsoft.AspNetCore.OData
+```
+
+---
+
+Internally, ASP.NET Core OData depends on the following.NET libraries for OData: 
+- [`Microsoft.OData.Core`](https://www.nuget.org/packages/Microsoft.OData.Core), which provides features like reading and writing OData payloads, parsing OData URIs and query options
+- [`Microsoft.OData.Edm`](https://www.nuget.org/packages/Microsoft.OData.Edm), which provides support for working with OData schemas, also referred to as EDM (Entity Data Model) model.
+
```
ElizabethOkerio, line 41: how come you didn't list the spatial library here.

habbes: I didn't think it was significant enough to be mentioned. But I can add it.

ElizabethOkerio (approved):

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+ms.author: clhabins
+
+ms.date: 1/5/2023
+---
+# Fundamentals Overview
+**Applies To**:[!INCLUDE[appliesto-webapi](../../includes/appliesto-webapi-v8.md)]
+
+This article provides an overview of the fundamental concepts for building ASP.NET Core OData services.
+
+ASP.NET Core OData is a .NET library that sits on top of ASP.NET Core to help you build OData REST APIs. The ASP.NET Core OData library consists of a set of services and middleware that hook into ASP.NET Core's request pipeline to provide features like routing, query handling and serialization based on OData specifications and conventions. All these concepts will be covered in depth in separate articles, this just provides an overview.
```
gathogojr, line 15:
```suggestion
ASP.NET Core OData is a .NET library built on top of ASP.NET Core to help you build OData REST APIs. The ASP.NET Core OData library consists of a set of services and middleware that hook into ASP.NET Core's request pipeline to provide features like routing, query handling and serialization based on OData specifications and conventions. All these concepts will be covered in depth in separate articles, this just provides an overview.
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+
+To build an ASP.NET Core OData app, you start with an ASP.NET Core application, then install the [`Microsoft.AspNetCore.OData`](https://www.nuget.org/packages/Microsoft.AspNetCore.OData) package as a dependency from NuGet. ASP.NET Core OData 8 supports .NET Core 3.1 and .NET Core 6.0 and above.
```
gathogojr, line 17:
```suggestion
To build an ASP.NET Core OData app, you start with an ASP.NET Core application, then install the [`Microsoft.AspNetCore.OData`](https://www.nuget.org/packages/Microsoft.AspNetCore.OData) Nuget package as a dependency. ASP.NET Core OData 8 supports .NET Core 3.1 and .NET Core 6.0 and above.
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+
+var builder = WebApplication.CreateBuilder(args);
+
+// Create the service EDM model.
+var modelBuilder = new ODataConventionModelBuilder();
+modelBuilder.EntitySet<Customer>("Customers");
+var edmModel = modelBuilder.GetEdmModel();
+
+// Register OData service.
+services.AddControllers().AddOData(
```
gathogojr, line 63:
```suggestion
builder.Services.AddControllers().AddOData(
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+
+```c#
+services.AddControllers().AddOData(options => options.AddRouteComponents(edmModel));
+```
+
+There are a number of ways to create an `IEdmModel` instance:
+- You can use the `CsdlReader` class to parse an `IEdmModel` from a stream that contains the OData schema in [CSDL XML](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html) or [CSDL JSON](https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html) format. The `CsdlReader` class is defined in the `Microsoft.OData.Edm` package. To learn more, [visit this article](/odata/odatalib/edm/read-write-model).
```
gathogojr, line 148:
```suggestion
- Using the `CsdlReader` class to parse an `IEdmModel` from a stream that contains the OData schema in [CSDL XML](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html) or [CSDL JSON](https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html) format. The `CsdlReader` class is defined in the `Microsoft.OData.Edm` package. To learn more, [visit this article](/odata/odatalib/edm/read-write-model).
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+- You can manually construct a model by creating an instance of the [`EdmModel`](/dotnet/api/microsoft.odata.edm.edmmodel) class (which implements `IEdmModel`). The `EdmModel` class is also part of the `Microsoft.OData.Edm` package. To learn more about creating EDM models manually, visit the [EDM documentation](/odata/odatalib/edm/build-basic-model).
+- You can use the `Microsoft.OData.ModelBuilder` package to easily create an `IEdmModel` based on CLR classes. The package provides 2 main classes:
```
gathogojr, line 150:
```suggestion
- Using the `Microsoft.OData.ModelBuilder` package to easily create an `IEdmModel` based on CLR classes. The package provides 2 main classes:
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+
+```c#
+services.AddControllers().AddOData(options => options.AddRouteComponents(edmModel));
+```
+
+There are a number of ways to create an `IEdmModel` instance:
+- You can use the `CsdlReader` class to parse an `IEdmModel` from a stream that contains the OData schema in [CSDL XML](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html) or [CSDL JSON](https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html) format. The `CsdlReader` class is defined in the `Microsoft.OData.Edm` package. To learn more, [visit this article](/odata/odatalib/edm/read-write-model).
+- You can manually construct a model by creating an instance of the [`EdmModel`](/dotnet/api/microsoft.odata.edm.edmmodel) class (which implements `IEdmModel`). The `EdmModel` class is also part of the `Microsoft.OData.Edm` package. To learn more about creating EDM models manually, visit the [EDM documentation](/odata/odatalib/edm/build-basic-model).
```
gathogojr, line 149:
```suggestion
- Manually constructing a model by creating an instance of the [`EdmModel`](/dotnet/api/microsoft.odata.edm.edmmodel) class (which implements `IEdmModel`). The `EdmModel` class is also part of the `Microsoft.OData.Edm` package. To learn more about creating EDM models manually, visit the [EDM documentation](/odata/odatalib/edm/build-basic-model).
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+- You can use the `Microsoft.OData.ModelBuilder` package to easily create an `IEdmModel` based on CLR classes. The package provides 2 main classes:
+    - `ODataConventionModelBuilder` used to quickly create models from CLR classes based on opinionated conventions (for example, it can infer key properties based on conventional names like `Id`). This is the easiest method and what we use in most of the samples in this documentation. For example usage, [visit this article](/odata/webapi/convention-model-builder).
+    - `ODataModelBuilder` gives you more control to specify things explicitly but requires more work to use. For example usage, [visit this article](/odata/webapi/model-builder-nonconvention).
```
gathogojr, line 152: I think because you're talking about the 2 main classes, this sentence should probably start with `- `ODataModelBuilder` that gives you...`.

habbes: I've rephrased it as "`ODataModelBuilder`, that gives you..."

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+</edmx:Edmx>
+```
+
+Clients can consume this metadata to know what types and endpoints the service exposes. Many OData clients use this metadata to generate strongly-typed client code. Tools like the [OData Connected Service](/odata/connectedservice/getting-started) and [OData CLI](/odata/odatacli/getting-started) can generate client code based on an OData schema.
+
+## Registering OData services
+
+It is possible to service multiple OData APIs in the same ASP.NET Core application.
+
+The sample code in the introduction registers a single OData service using `options.AddRouteComponents(edmModel)`. This is hosted on the "default" route, i.e. on the root path `/`. This means that the endpoints on this OData service look like :
```
gathogojr, line 210: How would someone know where the introduction is? Maybe you could make this a section link... `The sample code in the [introduction](#introduction) section`. Though I don't see a section with "Introduction" as a title in the document

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+- `$filter` allows you to filter results based on specific conditions: `/Products?$filter=Price lt 1000 and Category eq 'Electronics'` will only return electronic products whose price is less than 1,000.
+- `$select` allows you to limit the fields that are included in the response: `/Products?$select=Name,Category,DateCreated` will only include the `Name`, `Category`, `DateCreated` fields in the response
+- `$expand` allows you to join and include related data in the response: `/OrderItems?$expand=Product` will return order items and include the related product in the response
+- `$skip` and `$top` allow you to limit the number of records in the response and can be used for client-driven pagination: `/Products$top=10&$skip=20` returns the third page of 10 products.
+- `$orderby` allows you to sort the results: `/Products?$orderby=DateCreated desc` will sort the products by date in descending order (most recent first).
+- `$apply` allows you to apply aggregations: `/Orders?$apply=groupby((Product/Name), aggregate($count as Count))` returns the number of orders per product name.
```
gathogojr, line 295:
```suggestion
- `$apply` allows you to group and aggregate results: `/Orders?$apply=groupby((Product/Name), aggregate($count as Count))` returns the number of orders per product name.
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+
+These query options are not enabled by default. You can enable and configure them individually through the options passed to the `AddOData()` method:
```
gathogojr, line 297:
```suggestion
These query options are not enabled by default. You can enable and configure them individually through the options parameter passed to the `AddOData()` method:
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+In this example, we only enable the `$select`, `$filter`, `$expand`, `$orderby` and `$top` query options and we're setting the max allowable value for `$top` to be 100. If the client tries to use an OData query option that's not enabled, the service will return an error.
+
+You can enable all supported query options in one go using the `EnableQueryFeatures` method:
+
+```c#
+services.AddControllers().AddOData(
+    options => options.EnableQueryFeatures(100).AddRouteComponents(edmModel));
+```
+
+The argument passed to `EnableQueryFeatures` is the max top value. You can set it to `null` if you don't want to impose a limit the value specified in the `$top` query option.
```
gathogojr, line 312: "to impose a limit the value specified in the `$top` query option" - fix grammar

Odata-docs/webapi-8/fundamentals/overview.md
habbes: I intended to create a variable `var services = builder.Services` but I think I forgot that part. The idea was to make the statement consistent with the .NET Core 3.1 sample so that it would be easier to refer to both in explanations. But I've removed references to `services.` in inline code snippets, I'm now referring to methods directly, based on your offline comments.

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+
+After enabling query options on the service, you also need to enable querying on specific endpoints where you want to support query options. You can do this by adding the [`[EnableQuery]`](/dotnet/api/microsoft.aspnetcore.odata.query.enablequeryattribute) attribute to the corresponding controller actions. The [`EnableQuery`] attribute automatically applies the query options to the results returned by the controller action.
```
gathogojr, line 314:
```suggestion
After enabling query options on the service, you also need to enable querying on specific endpoints where you want to support query options. You can do this by adding the [`EnableQuery`](/dotnet/api/microsoft.aspnetcore.odata.query.enablequeryattribute) attribute to the corresponding controller actions. The `EnableQuery` attribute automatically applies the query options to the results returned by the controller action.
```

I don't think you need to include the square brackets every time you're referring to a method attribute

Odata-docs/webapi-8/fundamentals/overview.md
habbes: Yeah, I initially said "above". I could add an introduction title. But what if I just said "at the beginning of this document"?

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+
+## Request and response format
+
+ASP.NET Core OData uses JSON format for its requests and responses (the `$metadata` endpoint returns the schema in XML format by default). The OData JSON format may include metadata in the payload such as type information that makes it possible for clients and server to know how to handle the payload. This metadata is usually expressed in the form of "annotations" which are extra fields in the payload that start with `@odata.` prefix.
+
+For example. Let's say we have an OData service that has a `Customers` entity set which returns entities of type `Customer`.
+
+For a request like `GET /Customers`, we would a get a response like
```
gathogojr, line 354:
```suggestion
For a request like `GET /Customers`, we would a get a response like:
```

Odata-docs/webapi-8/fundamentals/overview.md
habbes: Would it be clear that it's an attribute? Or should I call it `EnableQueryAttribute`?

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+        },
+        {
+            "Id": 3,
+            "Name": "Customer 3"
+        }
+    ]
+}
+```
+
+Notice that this response does not just return an array of `Customer` object, it returns an object comprising of `@odata.context` and `value` properties. The `value` property's value is the array that contains the actual data. The `@odata.context` property is an annotation that gives that client information about where this data comes from. The `http://localhost:5000/odata/$metadata#Customers` is called the context URL, in this case it shows that this payload comes from the `Customers` entity set. The client can find out more information about the `Customers` entity set from the schema, which it can be retrieved from the service metadata endpoint - `http://localhost:5000/$metadata`. Using both the schema and the context URL, the client can tell that the payload value is a an array of `Customer` objects, and it can determine the properties contained in the `Customer` entity type and their respective types.
```
gathogojr, line 376:
```suggestion
Notice that this response does not just return an array of `Customer` object, it returns an object comprising of `@odata.context` and `value` properties. The `value` property's value is the array that contains the actual data. The `@odata.context` property is an annotation that gives that client information about where this data comes from. The `http://localhost:5000/$metadata#Customers` is called the context URL, in this case it shows that this payload comes from the `Customers` entity set. The client can find out more information about the `Customers` entity set from the schema, which can be retrieved from the service metadata endpoint - `http://localhost:5000/$metadata`. Using both the schema and the context URL, the client can tell that the payload value is a an array of `Customer` objects, and it can determine the properties contained in the `Customer` entity type and their respective types.
```

Odata-docs/webapi-8/fundamentals/overview.md
habbes: Or, "our first sample code in this document...?"

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+            "Name": "Customer 2"
+        },
+        {
+            "Name": "Customer 3"
+        }
+    ]
+}
+```
+
+Notice that the context URL has changed. Now the last part looks like `Customers(Name)` which means that the payload is from the `Customers` entity set (and therefore of the `Customer` type), but only includes the `Name` field. So we see that the response type does not only depend on the types defined in the model, but also the data transformations based on the request. The context URL and other annotations help the client figure out how to handle such dynamic payloads in a predictable and type-safe manner.
```
gathogojr, line 399:
```suggestion
Notice that the context URL is different in this case. Now the last part looks like `Customers(Name)` which means that the payload is from the `Customers` entity set (and therefore of the `Customer` type), but only includes the `Name` field. So we see that the response type does not only depend on the types defined in the model, but also the data transformations based on the request. The context URL and other annotations help the client figure out how to handle such dynamic payloads in a predictable and type-safe manner.
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+
+## Batch requests
+
+OData allows you to group multiple operations in a single HTTP request, called a batch request. The response to this batch request contains the responses to the individual operations that were in the batch request. Batch requests allow you to reduce the number of round trips between client and server and can improve the performance and scalability of a service.
+
+ASP.NET Core OData supports OData batch requests, but they are not enabled by default. To enable batch requests, you need to provide a batch handler and enable the OData batch middleware. The batch handler is the service that's responsible for processing a batch request. The batch middleware intercepts batch requests and sends them to the batch handler instead of conventional routing. The library provides a default batch handler to make batch support easy.
```
gathogojr, line 407:
```suggestion
The ASP.NET Core OData batch requests feature is not enabled by default. To enable batch requests, you need to provide an OData batch handler and enable the OData batch middleware. The OData batch handler is the service that's responsible for processing a batch request. The OData batch middleware intercepts batch requests and routes them to the batch handler. The ASP.NET Core OData library provides a default batch handler to make batch support easy.
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+using Microsoft.OData.ModelBuilder;
+
+var builder = WebApplication.CreateBuilder(args);
+
+var modelBuilder = new ODataConventionModelBuilder();
+modelBuilder.EntitySet<Customer>("Customers");
+var edmModel = modelBuilder.GetEdmModel();
+
+// Register OData service with support for batch requests.
+services.AddControllers().AddOData(
```
gathogojr:
```suggestion
builder.Services.AddControllers().AddOData(
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+
+        app.UseRouting();
+        app.UseEndpoints(endpoints => endpoints.MapControllers());
+    }
+}
+```
+
+---
+
+We add an instance of the [`DefaultODataBatchHandler`](/dotnet/api/microsoft.aspnetcore.odata.batch.defaultodatabatchhandler) as an argument to the the `options.AddRouteComponents()` method. We register the batching middleware using `app.UseODataBatching()`.
```
gathogojr:
```suggestion
In the preceding block of code, we add an instance of the [`DefaultODataBatchHandler`](/dotnet/api/microsoft.aspnetcore.odata.batch.defaultodatabatchhandler) as an argument to the the `options.AddRouteComponents()` method. We also register the batching middleware using `app.UseODataBatching()`. 
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+```
+
+---
+
+We add an instance of the [`DefaultODataBatchHandler`](/dotnet/api/microsoft.aspnetcore.odata.batch.defaultodatabatchhandler) as an argument to the the `options.AddRouteComponents()` method. We register the batching middleware using `app.UseODataBatching()`. 
+
+> [!NOTE]
+> The order of middleware matters. It is important that `app.UseODataBatching()` is called before `app.UseRouting()`.
+
+With this setup, the service can now handle batch requests. OData batch requests are POST requests made to the `/$batch` endpoint. The request body contains the individual operations, these operations can independent or related `GET`, `PUT`, `PATCH`, `POST`, `DELETE` requests or function and action invocations. The body of the batch request supports two formats: JSON and multipart mixed. The format is specified in the `Content-Type` header: `application/json` for JSON and `multipart/mixed` for multipart mixed. The following sample shows an example of a JSON batch request:
```
gathogojr, line 478:
```suggestion
With this setup, the service can now handle batch requests. OData batch requests are POST requests made to the `/$batch` endpoint. The request body contains the individual operations. The operations can independent or related `GET`, `PUT`, `PATCH`, `POST`, `DELETE` requests or function and action invocations. The body of the batch request supports two formats: JSON and multipart mixed. The format is specified in the `Content-Type` header: `application/json` for JSON and `multipart/mixed` for multipart mixed. The following sample shows an example of a JSON batch request:
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+**Applies To**:[!INCLUDE[appliesto-webapi](../../includes/appliesto-webapi-v8.md)]
+
+This article provides an overview of the fundamental concepts for building ASP.NET Core OData services.
+
+ASP.NET Core OData is a .NET library built on top of ASP.NET Core to help you build OData REST APIs. The ASP.NET Core OData library consists of a set of services and middleware that hook into ASP.NET Core's request pipeline to provide features like routing, query handling and serialization based on OData specifications and conventions. All these concepts will be covered in depth in separate articles, this article just provides an overview.
+
+To build an ASP.NET Core OData app, you start with an ASP.NET Core application, then install the [`Microsoft.AspNetCore.OData`](https://www.nuget.org/packages/Microsoft.AspNetCore.OData) NuGet package as a dependency. ASP.NET Core OData 8 supports .NET Core 3.1 and .NET Core 6.0 and above.
+
+> [!NOTE]
+> ASP.NET COre OData 8 supports .NET Core 3.1 and .NET Core 6.0 and above.
```
gathogojr, line 20:
```suggestion
> ASP.NET Core OData 8 supports .NET Core 3.1 and .NET Core 6.0 and above.
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+
+# [.NET Core CLI](#tab/netcore-cli)
+
+```dotnetcli
+dotnet add package Microsoft.AspNetCore.OData
+```
+
+---
+
+Internally, ASP.NET Core OData depends on the following.NET libraries for OData:
```
gathogojr, line 38:
```suggestion
Internally, ASP.NET Core OData depends on the following .NET libraries for OData: 
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+# Fundamentals Overview
+**Applies To**:[!INCLUDE[appliesto-webapi](../../includes/appliesto-webapi-v8.md)]
+
+This article provides an overview of the fundamental concepts for building ASP.NET Core OData services.
+
+ASP.NET Core OData is a .NET library built on top of ASP.NET Core to help you build OData REST APIs. The ASP.NET Core OData library consists of a set of services and middleware that hook into ASP.NET Core's request pipeline to provide features like routing, query handling and serialization based on OData specifications and conventions. All these concepts will be covered in depth in separate articles, this article just provides an overview.
+
+To build an ASP.NET Core OData app, you start with an ASP.NET Core application, then install the [`Microsoft.AspNetCore.OData`](https://www.nuget.org/packages/Microsoft.AspNetCore.OData) NuGet package as a dependency. ASP.NET Core OData 8 supports .NET Core 3.1 and .NET Core 6.0 and above.
+
+> [!NOTE]
```
gathogojr, line 19: When I saw this note, I thought it's used to emphasize a statement from the body of the doc. Then I went further down and came across a note that doesn't reference a statement from the body of the doc. What is the practice?

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+    - `ODataModelBuilder`, that gives you more control to specify things explicitly but requires more work to use. For example usage, [visit this article](/odata/webapi/model-builder-nonconvention).
+
+> [!NOTE]
+> In ASP.NET Core OData 7, the model builder was built-in to the library. It was extracted into a standalone package `Microsoft.OData.Modelbuilder`. The two work the same mostly and the 7.x model builder documentation still applies to the standalone `Microsoft.OData.ModelBuilder`.
+
+Here are examples of things that can be defined in an EDM model:
+- Entity types
+- Complex types
+- Enums
+- Entity Sets
```
gathogojr, line 161:
```suggestion
- Entity sets
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+    {
+        app.UseRouting();
+        app.UseEndpoints(endpoints => endpoints.MapControllers());
+    }
+}
+```
+
+---
+
+We register OData services using the `AddOData` extension method. The `AddOData` method accepts a callback action that allows us to configure different aspects of the OData service using the options argument (which is an instance of [`ODataOptions`](/dotnet/api/microsoft.aspnetcore.odata.odataoptions)).
```
gathogojr, line 107: Maybe add a link to the `AddOData` method just like you have done for `ODataOptions` and `AddRouteComponents`

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+</edmx:Edmx>
+```
+
+Clients can consume this metadata to know what types and endpoints the service exposes. Many OData clients use this metadata to generate strongly-typed client code. Tools like the [OData Connected Service](/odata/connectedservice/getting-started) and [OData CLI](/odata/odatacli/getting-started) can generate client code based on an OData schema.
+
+## Registering OData services
+
+It is possible to service multiple OData APIs in the same ASP.NET Core application.
+
+The first sample code in this document registers a single OData service using `AddRouteComponents(edmModel)`. This is hosted on the "default" route, i.e. on the root path `/`. This means that the endpoints on this OData service look like :
```
gathogojr, line 210:
```suggestion
The first sample code in this document registers a single OData service using `AddRouteComponents(IEdmModel)`. This is hosted on the "default" route, i.e. on the root path `/`. This means that the endpoints on this OData service look like :
```
The way to represent a method that accepts an argument is by including the argument type...

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+```c#
+AddOdata(
+    options => options.AddRouteComponents("odata", edmModel));
+```
+In this case, paths will be prefixed with `odata`:
+- `/odata/Customers`, 
+- `/odata/Products`, 
+- `/odata/$metadata`
+
+The application's base URL together with the route prefix form the **service root** of the OData service. For example, if the application's base URL is `https://service.example.org` and the route prefix is `odata`, then the service root for this OData service will be `https://service.example.org/odata`. If the route prefix is empty (i.e. default route), then the service root will be `https://service.example.org`.
```
gathogojr, line 226:
```suggestion
The application's base URL together with the route prefix form the **service root** of the OData service. For example, if the application's base URL is `https://service.example.org` and the route prefix is `odata`, then the service root for the OData service will be `https://service.example.org/odata`. If the route prefix is empty (i.e. default route), then the service root will be `https://service.example.org`.
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+```
+
+This will create OData services with the following url:
+- `<baseUrl>/api1` using the schema defined by `edmModel1`.
+- `<baseUrl>/api2` using the schema defined by `edmModel2`.
+
+## Routing
+
+> [!IMPORTANT]
+> ASP.NET Core OData routing is based on controllers and therefore does not support [minimal APIs](/aspnet/core/fundamentals/minimal-apis/overview).
```
gathogojr, line 242:
```suggestion
> ASP.NET Core OData routing is based on controllers and currently does not support [minimal APIs](/aspnet/core/fundamentals/minimal-apis/overview).
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+    {
+        return Ok(_db.Customers);
+    }
+
+    public ActionResult<Customer> Get(int id)
+    {
+        return Ok(_db.Customers.Find(id));
+    }
+
+    public ActionResult<Customer> Post(Customer customer)
```
gathogojr, line 268:
```suggestion
    public ActionResult<Customer> Post([FromBody] Customer customer)
```
`FromBody` attribute is required if the payload is in the request body

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+    {
+        _db.Customers.Add(customer);
+        _db.SaveChanges();
+        return Created(customer);
+    }
+}
+```
+
+- The `Get()` controller action will be mapped to the `GET /Customers` endpoint.
+- The `Get(int id)` controller action will be mapped to the `GET /Customers({id})` and `GET /Customers/{id}` endpoints
```
gathogojr, line 278:
```suggestion
- The `Get(int id)` controller action will be mapped to the `GET /Customers({id})` and `GET /Customers/{id}` endpoints.
```

Odata-docs/webapi-8/fundamentals/overview.md
```diff
@@ -0,0 +1,582 @@
...
+In this example, we only enable the `$select`, `$filter`, `$expand`, `$orderby` and `$top` query options and we're setting the max allowable value for `$top` to be 100. If the client tries to use an OData query option that's not enabled, the service will return an error.
+
+You can enable all supported query options in one go using the `EnableQueryFeatures` method:
+
+```c#
+AddOData(
+    options => options.EnableQueryFeatures(100).AddRouteComponents(edmModel));
+```
+
+The argument passed to `EnableQueryFeatures` is the max top value. You can set it to `null` if you don't want to impose a limit to the value specified in the `$top` query option.
```
gathogojr, line 312:
```suggestion
The argument passed to `EnableQueryFeatures` is the max allowable value for `$top` query option. You can set it to `null` if you don't want to impose a limit to the value specified in the `$top` query option.
```

gathogojr (approved):
Left a few comments, otherwise LGTM

Odata-docs/webapi-8/fundamentals/overview.md
habbes: This is the use case from the NOTE alert according to the [guidelines](https://learn.microsoft.com/en-us/contribute/markdown-reference#alerts-note-tip-important-caution-warning): Information the user should notice even if skimming.

My aim was to emphasize something that I thought was important, especially important "gotchas" that could affect the developer experience if not taken into account. But the guidelines discourages frequent use of alerts within the same docs:

> Avoid notes, tips, and important boxes. Readers tend to skip over them. It's better to put that info directly into the article text.

> If you need to use alerts, limit them to one or two per article. Multiple notes should never be next to each other in an article.

So let me remove the note since it's already referenced in the docs.

Odata-docs/webapi-8/fundamentals/overview.md
habbes: I've rephrased it as "maximum allowed value for `$top` query option"

PR #281, opened 2023-01-12
ASP.NET Core OData 8 pagination docs

habbes:
ASP.NET Core OData 8 pagination docs, consisting of:
- Pagination overview
- Client-driven paging
- Server-driven paging

A good follow up to this would be a walkthrough/guide on creating a custom `SkipTokenHandler`.

### DIFF
--- a/Odata-docs/TOC.yml
+++ b/Odata-docs/TOC.yml
@@ -79,10 +79,10 @@
                 href: /odata/webapi-8/fundamentals/ref-routing
           - name: Actions and Functions
             href: /odata/webapi-8/fundamentals/actions-functions
-          - name: Pagination
+          - name: Paging
             items:
-              - name: Pagination overview
-                href: /odata/webapi-8/fundamentals/pagination-overview
+              - name: Paging overview
+                href: /odata/webapi-8/fundamentals/paging-overview
               - name: Client-driven paging
                 href: /odata/webapi-8/fundamentals/client-driven-paging
               - name: Server-driven paging
--- a/Odata-docs/TOC.yml
+++ b/Odata-docs/TOC.yml
@@ -79,6 +79,14 @@
                 href: /odata/webapi-8/fundamentals/ref-routing
           - name: Actions and Functions
             href: /odata/webapi-8/fundamentals/actions-functions
+          - name: Pagination
+            items:
+              - name: Pagination overview
+                href: /odata/webapi-8/fundamentals/pagination-overview
+              - name: Client-driven paging
+                href: /odata/webapi-8/fundamentals/client-driven-paging
+              - name: Server-driven paging
+                href: /odata/webapi-8/fundamentals/server-driven-paging
       - name: Tutorials
         items:
           - name: Basic CRUD operations
--- a/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
+++ b/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
@@ -122,6 +122,8 @@ This returns the following result:
 
 Notice the client did not see product 3 because it was removed before it fetched page 2. In a similar way, if new products were added to the data source before product 2, then the client would see product 2 again in page 2 or later pages.
 
+If the underlying data source collection is not ordered, you may also need to manually order the results to have stable pages, e.g. `GET /Products?$top=2&$skip=2&$orderby=Id`.
+
 Using `$skip` can also lead to performance issues in large datasets in some data source implementation. In many SQL-based database systems `$skip` is implemented using `OFFSET`. A request like `/Products?$skip=50000&$top=100` could be translated in the following SQL query:
 
 ```sql
--- a/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
+++ b/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
@@ -146,7 +146,7 @@ LIMIT 100, OFFSET 50000;
 
 This query will return 100 rows, from row 50,001 to row 50,100. However it will scan through all the first 50,100 rows in the table, then drop the first 50,000. The larger the offset, the longer the query will take.
 
-This makes `$skip`-based pagination unsuitable for large datasets in many database systems. An alternative approach would be to keep track of where you left off using a unique ordered field. In our example, we can use the `Id` field to keep track of the last product we saw. using this technique, the client can fetch the first page with the following request:
+This makes `$skip`-based paging unsuitable for large datasets in many database systems. An alternative approach would be to keep track of where you left off using a unique ordered field. In our example, we can use the `Id` field to keep track of the last product we saw. using this technique, the client can fetch the first page with the following request:
 
 ```http
 GET /Products?$top=2
--- a/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
+++ b/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
@@ -8,11 +8,13 @@ ms.author: clhabins
 ---
 
 # Client-driven paging in ASP.NET Core OData 8
+
 **Applies To**:[!INCLUDE[appliesto-webapi](../../includes/appliesto-webapi-v8.md)]
 
 The client can request the OData service to return a specific number of results. The `$top` query option is used to limit the number of results and the `skip` query option used to specify the offset. For example `GET /Customers?$skip=3$top=5` will return up to 5 items starting from the 4th item in the collection (i.e. after skipping the first 3).
 
 Using these two query options the client can divide a large collection into separate pages and fetch the pages in separate requests. For example, if our collection has 25 items, and we want each page to have 10 items, then we can fetch the different page as follows:
+
 - First page: `$top=10`
 - Second page: `$skip=10&$top=10`
 - Third page: `$skip=20&top=10` (The last page will return only 5 items)
@@ -63,6 +65,7 @@ public class ProductsController : ODataController
 ## Benefits of this approach
 
 This client-driven paging approach has a number of benefits:
+
 - It's relatively easy to implement
 - It allows the client to fetch pages independently and to jump to an arbitrary page. For example, the client can fetch page 5 without having fetched page 4 previously.
 - If the client knows the total size of the collection (e.g. using the `$count` query option), the client can compute the total number of pages as well as generate links for each page in advance.
@@ -72,6 +75,7 @@ This client-driven paging approach has a number of benefits:
 Paging based on `$top` and `$skip` has a few drawbacks that may make it unsuitable in some scenarios.
 
 This approach is sensitive to changes in that data source which may lead the page contents to be unstable. To illustrate this, let's assume we have a data source with the following 5 items and we want to fetch 2 items per page:
+
 ```json
 [
     { "Id": 1, "Name": "Product 1" },
@@ -81,11 +85,13 @@ This approach is sensitive to changes in that data source which may lead the pag
     { "Id": 5, "Name": "Product 5" }
 ]
 ```
+
 The client can fetch the first page with 
 
 ```http
 GET /Products?$top=2
 ```
+
 This returns the following result:
 
 ```json
@@ -99,6 +105,7 @@ This returns the following result:
 ```
 
 Let's say that product 2 is removed from the data source before the client fetches the second page. The data source looks like:
+
 ```json
 [
     { "Id": 1, "Name": "Product 1" },
@@ -136,6 +143,7 @@ Using `$skip` can also lead to performance issues in large datasets in some data
 SELECT * FROM products
 LIMIT 100, OFFSET 50000;
 ```
+
 This query will return 100 rows, from row 50,001 to row 50,100. However it will scan through all the first 50,100 rows in the table, then drop the first 50,000. The larger the offset, the longer the query will take.
 
 This makes `$skip`-based pagination unsuitable for large datasets in many database systems. An alternative approach would be to keep track of where you left off using a unique ordered field. In our example, we can use the `Id` field to keep track of the last product we saw. using this technique, the client can fetch the first page with the following request:
@@ -143,6 +151,7 @@ This makes `$skip`-based pagination unsuitable for large datasets in many databa
 ```http
 GET /Products?$top=2
 ```
+
 This will return the following result:
 
 ```json
@@ -154,18 +163,23 @@ This will return the following result:
     ]
 }
 ```
+
 When the client gets this result, it records that the last product it saw was the product with ID 2. So, to fetch the next page, it makes the following request:
+
 ```http
 GET /Products?$filter=Id gt 2&$top=2
 ```
+
 This fetches only products with id greater than 2.
 
 This roughly translate to the following SQL query:
+
 ```sql
 SELECT * FROM products
 WHERE id > 2
 LIMIT 2;
 ```
+
 If the database has an ordered index on the `id` column, then this query will scale well and perform significantly better on larger datasets and pages than if it was using `OFFSET`.
 
 This approach also makes the pages more resilient to concurrent changes in the data source. But we lose some features like being able to easily compute the total number of pages or to jump to a random page. We are limited to visiting the pages sequentially.
--- a/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
+++ b/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
@@ -89,10 +89,13 @@ GET /Products?$top=2
 This returns the following result:
 
 ```json
-[
-    { "Id": 1, "Name": "Product 1" },
-    { "Id": 2, "Name": "Product 2" }
-]
+{
+    "@odata.context": "http://localhost:5000/$metadata#Products",
+    "value": [
+        { "Id": 1, "Name": "Product 1" },
+        { "Id": 2, "Name": "Product 2" }
+    ]
+}
 ```
 
 Let's say that product 2 is removed from the data source before the client fetches the second page. The data source looks like:
@@ -114,10 +117,13 @@ GET /Products?$top=2&$skip=2
 This returns the following result:
 
 ```json
-[
-    { "Id": 4, "Name": "Product 4" },
-    { "Id": 5, "Name": "Product 5" }
-]
+{
+    "@odata.context": "http://localhost:5000/$metadata#Products",
+    "value": [
+        { "Id": 4, "Name": "Product 4" },
+        { "Id": 5, "Name": "Product 5" }
+    ]
+}
 ```
 
 Notice the client did not see product 3 because it was removed before it fetched page 2. In a similar way, if new products were added to the data source before product 2, then the client would see product 2 again in page 2 or later pages.
@@ -140,10 +146,13 @@ GET /Products?$top=2
 This will return the following result:
 
 ```json
-[
-    { "Id": 1, "Name": "Product 1" },
-    { "Id": 2, "Name": "Product 2" }
-]
+{
+    "@odata.context": "http://localhost:5000/$metadata#Products",
+    "value": [
+        { "Id": 1, "Name": "Product 1" },
+        { "Id": 2, "Name": "Product 2" }
+    ]
+}
 ```
 When the client gets this ersult, it records that the last product it saw was the product with ID 2. So, to fetch the next page, it makes the following request:
 ```http
--- a/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
+++ b/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
@@ -33,7 +33,7 @@ By default, ASP.NET Core OData 8 limits the maximum value for `$top` to 0. So if
 
 Configuring client-side paging involves two steps:
 
-1. Configuring the max `$top` query option value
+1. Configuring the`$top` query option with a max value
 2. Adding `EnableQuery` attribute to the controller method
 
 You can configure the maximum top value using the [`ODataOptions.SetMaxTop()`](/dotnet/api/microsoft.aspnetcore.odata.odataoptions.setmaxtop) when adding OData services to your application. The following snippet would set the maximum allowed `$top` value to 100.
@@ -47,7 +47,7 @@ services.AddOData(options =>
 
 Now this will support values of up to 100, but return an error if `$top` exceeds 100.
 
-You can also remove the limit by setting it to `null`:
+You can also remove the limit by setting it to `null`. This allows `$top` to be set to any value greater or equal to 0:
 
 ```csharp
 services.AddOData(options =>
--- a/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
+++ b/Odata-docs/webapi-8/fundamentals/client-driven-paging.md
@@ -60,7 +60,103 @@ public class ProductsController : ODataController
 }
 ```
 
+## Benefits of this approach
+
 This client-driven paging approach has a number of benefits:
 - It's relatively easy to implement
-- It allows the client to fetch pages independently. For example, the client can fetch page 5 without having fetched page 4 previously.
-- If the client knows the total size of the collection (e.g. using the `$count` query option), the client can compute all the available page numbers and their offset ranges in advance.
+- It allows the client to fetch pages independently and to jump to an arbitrary page. For example, the client can fetch page 5 without having fetched page 4 previously.
+- If the client knows the total size of the collection (e.g. using the `$count` query option), the client can compute the total number of pages as well as generate links for each page in advance.
+
+## Drawbacks
+
+Paging based on `$top` and `$skip` has a number of drawbacks that may make unsuitable in some scenarios.
+
+This approach is sensitive to changes in that data source which may lead the page contents to be unstable. To illustrate this, let's assume we have a data source with the following 5 items and we want to fetch 2 items per page:
+```json
+[
+    { "Id": 1, "Name": "Product 1" },
+    { "Id": 2, "Name": "Product 2" },
+    { "Id": 3, "Name": "Product 3" },
+    { "Id": 4, "Name": "Product 4" },
+    { "Id": 5, "Name": "Product 5" }
+]
+```
+The client can fetch the first page with 
+
+```http
+GET /Products?$top=2
+```
+This returns the following result:
+
+```json
+[
+    { "Id": 1, "Name": "Product 1" },
+    { "Id": 2, "Name": "Product 2" }
+]
+```
+
+Let's say that product 2 is removed from the data source before the client fetches the second page. The data source looks like:
+```json
+[
+    { "Id": 1, "Name": "Product 1" },
+    { "Id": 3, "Name": "Product 3" },
+    { "Id": 4, "Name": "Product 4" },
+    { "Id": 5, "Name": "Product 5" }
+]
+```
+
+The client can fetch the second page with
+
+```http
+GET /Products?$top=2&$skip=2
+```
+
+This returns the following result:
+
+```json
+[
+    { "Id": 4, "Name": "Product 4" },
+    { "Id": 5, "Name": "Product 5" }
+]
+```
+
+Notice the client did not see product 3 because it was removed before it fetched page 2. In a similar way, if new products were added to the data source before product 2, then the client would see product 2 again in page 2 or later pages.
+
+Using `$skip` can also lead to performance issues in large datasets in some data source implementation. In many SQL-based database systems `$skip` is implemented using `OFFSET`. A request like `/Products?$skip=50000&$top=100` could be translated in the following SQL query:
+
+```sql
+SELECT * FROM products
+LIMIT 100, OFFSET 50000;
+```
+This query will return 100 rows, from row 50,001 to row 50,100. However it will scan through all the first 50,100 rows in the table, then drop the first 50,000. The larger the offset, the longer the query will take.
+
+This makes `$skip`-based pagination unsuitable for large datasets in many database systems. An alternative approach would be to keep track of where you left off using a unique ordered field. In our example, we can use the `Id` field to keep track of the last product we saw. using this technique, the client can fetch the first page with the following request:
+
+```http
+GET /Products?$top=2
+```
+This will return the following result:
+
+```json
+[
+    { "Id": 1, "Name": "Product 1" },
+    { "Id": 2, "Name": "Product 2" }
+]
+```
+When the client gets this ersult, it records that the last product it saw was the product with ID 2. So, to fetch the next page, it makes the following request:
+```http
+GET /Products?$filter=Id gt 2&$top=2
+```
+This fetches only products with id greater than 2.
+
+This roughly translate to the following SQL query:
+```sql
+SELECT * FROM products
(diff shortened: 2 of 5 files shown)


Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,173 @@
...
+## Benefits of this approach
+
+This client-driven paging approach has a number of benefits:
+- It's relatively easy to implement
+- It allows the client to fetch pages independently and to jump to an arbitrary page. For example, the client can fetch page 5 without having fetched page 4 previously.
+- If the client knows the total size of the collection (e.g. using the `$count` query option), the client can compute the total number of pages as well as generate links for each page in advance.
+
+## Drawbacks
+
+Paging based on `$top` and `$skip` has a number of drawbacks that may make unsuitable in some scenarios.
```
odero, line 72:
```suggestion
Paging based on `$top` and `$skip` has a few drawbacks that may make it unsuitable in some scenarios.
```

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,173 @@
...
+
+If the underlying data source collection is not ordered, you may also need to manually order the results to have stable pages, e.g. `GET /Products?$top=2&$skip=2&$orderby=Id`.
+
+Using `$skip` can also lead to performance issues in large datasets in some data source implementation. In many SQL-based database systems `$skip` is implemented using `OFFSET`. A request like `/Products?$skip=50000&$top=100` could be translated in the following SQL query:
+
+```sql
+SELECT * FROM products
+LIMIT 100, OFFSET 50000;
+```
+This query will return 100 rows, from row 50,001 to row 50,100. However it will scan through all the first 50,100 rows in the table, then drop the first 50,000. The larger the offset, the longer the query will take.
```
odero, line 139: This explains why `skip` isn't suitable but I was looking (here and in server-side doc) for the explanation of how `skipToken` works internally so that it doesn't suffer the same issues.

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,173 @@
...
+```json
+{
+    "@odata.context": "http://localhost:5000/$metadata#Products",
+    "value": [
+        { "Id": 1, "Name": "Product 1" },
+        { "Id": 2, "Name": "Product 2" }
+    ]
+}
+```
+When the client gets this ersult, it records that the last product it saw was the product with ID 2. So, to fetch the next page, it makes the following request:
```
odero, line 157:
```suggestion
When the client gets this result, it records that the last product it saw was the product with ID 2. So, to fetch the next page, it makes the following request:
```

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,173 @@
...
+```
+This fetches only products with id greater than 2.
+
+This roughly translate to the following SQL query:
+```sql
+SELECT * FROM products
+WHERE id > 2
+LIMIT 2;
+```
+If the database has an ordered index on the `id` column, the this query will scale well and perform significantly better on larger datasets and pages than if it wa using `OFFSET`.
```
odero, line 169:
```suggestion
If the database has an ordered index on the `id` column, then this query will scale well and perform significantly better on larger datasets and pages than if it was using `OFFSET`.
```

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,173 @@
...
+
+This approach also makes the pages more resilient to concurrent changes in the data source. But we lose some features like being able to easily compute the total number pages or to jump to a random page. We are limited to visit the pages sequentially.
```
odero, line 171:
```suggestion
This approach also makes the pages more resilient to concurrent changes in the data source. But we lose some features like being able to easily compute the total number of pages or to jump to a random page. We are limited to visiting the pages sequentially.
```

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,251 @@
...
+            "Name": "Product 2" 
+        },
+    ],
+    "@odata.nextLink": "http://localhost:5000/odata/Products?$skiptoken=foobar"
+}
+```
+
+In this example, the client would make a request to `GET http://localhost:5000/odata/Products?$skiptoken=foobar` to retrieve the next page.
+
+The format of the next link is opaque to the client. This means that the client should not try to decode the link, modify it or make any assumptions about how it is generated or what its components mean. It can only assume that it's a valid URL that will return the next page. This allows the server freedom in how it chooses to generate and interpret the next link. It allows the server to change the underlying implementation of the next link without affecting the client.
```
odero, line 36: The client-driven paging doc explains why internally, `skip` isn't suitable but I was looking here for the explanation of how `skipToken` works internally so that it doesn't suffer the same issues.

Odata-docs/TOC.yml
```diff
@@ -79,6 +79,14 @@
                 href: /odata/webapi-8/fundamentals/ref-routing
           - name: Actions and Functions
             href: /odata/webapi-8/fundamentals/actions-functions
+          - name: Pagination
+            items:
+              - name: Pagination overview
```
odero, line 84: nit: For consistency in the docs, consider sticking to either paging or pagination instead of both

Odata-docs/TOC.yml
```diff
@@ -79,6 +79,14 @@
                 href: /odata/webapi-8/fundamentals/ref-routing
           - name: Actions and Functions
             href: /odata/webapi-8/fundamentals/actions-functions
+          - name: Pagination
```
gathogojr, line 82: Given @odero's comment elsewhere maybe swap "Pagination" with "Paging" on the TOC
Pagination => Paging
Pagination overview = Paging overview

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,173 @@
...
+date:   2023-01-10
+ms.date: 1/10/2023
+author: habbes
+ms.author: clhabins
+---
+
+# Client-driven paging in ASP.NET Core OData 8
+**Applies To**:[!INCLUDE[appliesto-webapi](../../includes/appliesto-webapi-v8.md)]
+
+The client can implement paging by telling the OData service how many items to return for a given request. The `$top` query option is used to limit the number of results and the `skip` query option used to specify the offset. For example `GET /Customers?$skip=3$top=5` will return up to 5 items starting from the 4th item in the collection (i.e. after skipping the first 3).
```
gathogojr, line 13: The words "_client can implement paging_" do not do it for me... Especially the word "implement". Maybe "_The client can request the OData service to return a specific number of results_"

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,173 @@
...
+{
+    "error": {
+        "code": "",
+        "message": "The query specified in the URI is not valid. The limit of '0' for Top query has been exceeded. The value from the incoming request is '2'."
+}
+```
+
+You can configure the maximum top value using the [`ODataOptions.SetMaxTop()`](/dotnet/api/microsoft.aspnetcore.odata.odataoptions.setmaxtop) when adding OData services to your application. The following snippet would set the maximum allowed `$top` value to 100.
+
+```c#
```
gathogojr, line 32: I just noticed that \`\`\`c# renders a little different from \`\`\`csharp

\`\`\`c#:
image

vs

\`\`\`csharp:
image

Because of this, I think I'd suggest that you use \`\`\`csharp given that even \`\`\`json is rendered as:
image

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,251 @@
...
+        {
+            "Id": 1,
+            "Name": "Product 1" 
+        },
+        {
+            "Id": 2,
+            "Name": "Product 2" 
+        },
+    ],
+    "@odata.nextLink": "http://localhost:5000/odata/Products?$skiptoken=foobar"
```
gathogojr, line 30:
```suggestion
    "@odata.nextLink": "http://localhost:5000/odata/Products?$skiptoken=Id-2"
```
To make this relatable with the actual value for skip token. `skiptoken=Id-2` simply means the last item in the returned result had an `Id` value of 2.

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,251 @@
...
+        {
+            "Id": 1,
+            "Name": "Product 1"
+        },
+        {
+            "Id": 2,
+            "Name": "Product 2"
+        }
+    ],
+    "@odata.nextLink": "http://localhost:5000/Products?$skip=2"
```
gathogojr, line 76:
```suggestion
    "@odata.nextLink": "http://localhost:5000/Products?$skiptoken=Id-2"
```

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,251 @@
...
+        {
+            "Id": 3,
+            "Name": "Product 3"
+        },
+        {
+            "Id": 4,
+            "Name": "Product 4"
+        }
+    ],
+    "@odata.nextLink": "http://localhost:5000/Products?$skip=4"
```
gathogojr, line 99:
```suggestion
    "@odata.nextLink": "http://localhost:5000/Products?$skiptoken=Id-4"
```

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,251 @@
...
+
+If the request contains query options, they will also be included in the generated next link. Let's take the following request for example:
+
+```http
+GET http://localhost:5000/Products?$filter=Price gt 1000&$select=Name
+```
+
+The generated next link will look like:
+```
+http://localhost:5000/Products?$filter=Price%20gt%201000&$select=Name&$skip=2
```
gathogojr, line 129:
```suggestion
http://localhost:5000/Products?$filter=Price%20gt%201000&$select=Name&$skiptoken=Id-2
```

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,251 @@
...
+        {
+            "Id": 1,
+            "Name": "Product 1"
+        },
+        {
+            "Id": 2,
+            "Name": "Product 2"
+        }
+    ],
+    "@odata.nextLink": "http://localhost:5000/Products?$top=1&$skip=2"
```
gathogojr, line 158:
```suggestion
    "@odata.nextLink": "http://localhost:5000/Products?$top=1&$skiptoken=Id-2"
```

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,251 @@
...
+}
+```
+The response contains 2 products and a next link to the next page. The client fetches the next page using the next link:
+
+```http
+GET http://localhost:5000/Products?$top=1&$skip=2
```
gathogojr, line 164:
```suggestion
GET http://localhost:5000/Products?$top=1&$skiptoken=Id-2
```

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,251 @@
...
+            "Name": "Product 3"
+        }
+    ]
+}
+```
+This is now the last page. The server returns only one item and no next link.
+
+## Improving paging with `$skiptoken`
+
+You may have noticed that ASP.NET Core OData generates next links using `$skip`. Using `$skip` on large collections may lead to performance degradation in some databases. For more information about the drawbaks of `$skip`, visit the [client-driven paging](/odata/webapi-8/fundamentals/client-driven-paging#drawbacks) article.
```
gathogojr, line 182: I don't understand this paragraph. I expected that since this article talks about server-side paging, it'd focus on enabling skip token feature (`options.SkipToken()`) and applying `PageSize` configuration. My understanding was that those are the two cogs that drive server-side paging. However, what you're saying here is that `$skip` can be used to support server-side paging. That I wasn't aware. I might need to refresh my knowledge. I actually expected this article mostly on `$skiptoken` rather than `$skip`. It's also the reason why I left a lot of review suggestions to "fix" the next link

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,251 @@
...
+        {
+            "Id": 4,
+            "Name": "Product 4"
+        }
+    ],
+    "@odata.nextLink": "http://localhost:64771/Products?$skiptoken=Id-4"
+}
+```
+
+The `$skiptoken` value is generated using an implementation of [`SkipTokenHandler`](/dotnet/api/microsoft.aspnetcore.odata.query.skiptokenhandler). By default, the built-in [`DefaultSkipTokenHandler`](/dotnet/api/microsoft.aspnetcore.odata.query.defaultskiptokenhandler) class. `DefaultSkipTokenHandler` generates the `$skiptoken` based on the values of the key fields and fields in the `$orderby` query option if present.
```
gathogojr, line 240:
```suggestion
The `$skiptoken` value is generated using an implementation of [`SkipTokenHandler`](/dotnet/api/microsoft.aspnetcore.odata.query.skiptokenhandler), by default the built-in [`DefaultSkipTokenHandler`](/dotnet/api/microsoft.aspnetcore.odata.query.defaultskiptokenhandler) class. `DefaultSkipTokenHandler` generates the `$skiptoken` based on the values of the key fields and fields in the `$orderby` query option if present.
```

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,251 @@
...
+
+You can customize the `$skiptoken` by providing your own implementation of `SkipTokenHandler` and adding it to the OData route's service container:
```
gathogojr, line 242:
```suggestion
You can customize the `$skiptoken` by providing your own implementation of `SkipTokenHandler` and injecting it into the OData service dependency injection (DI) container:
```

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
habbes: What I meant by that is the fact that on the client-side you can have paging UI where the user can select a specific page number, or next/prev pages, and the logic that manages this, to keep track of pages, etc. lies on the client side.

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
habbes: You mean the language on the title is uppercase (C#) when using `csharp` and lowercase (c#) when using `c#`? I never noticed that before. Thanks.

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: This is an implementation detail. I wanted to show that the value of `$skiptoken` could be anything and the client shouldn't try to make sense of it.

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: WebAPI uses `$skip` in the next link by default if you don't explicitly enable `SkipToken()`. I mention this in the "Improving pagint with `$skiptoken` section:


> You may have noticed that ASP.NET Core OData generates next links using `$skip`. Using `$skip` on large collections may lead to performance degradation in some databases. For more information about the drawbaks of `$skip`, visit the [client-driven paging](/odata/webapi-8/fundamentals/client-driven-paging#drawbacks) article.

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: See: https://github.com/MicrosoftDocs/OData-docs/pull/281#discussion_r1087433123

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: See: https://github.com/MicrosoftDocs/OData-docs/pull/281#discussion_r1087433123

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: See: https://github.com/MicrosoftDocs/OData-docs/pull/281#discussion_r1087433123

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: See: https://github.com/MicrosoftDocs/OData-docs/pull/281#discussion_r1087433123

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: The way I understand it, server-side paging means relying on the server-generated next link. The next link can be generated whether or not you use `$skiptoken`. The next link is computed when you set the `PageSize` property. Skip tokens provide an extra layer of improvements on top of next link. It's also something I discovered while writing this docq when I tried things out and referencing older docs we have on the subject:
- https://learn.microsoft.com/en-us/odata/webapi/pagesize-top-maxtop
- https://learn.microsoft.com/en-us/odata/webapi/skipToken-for-server-side-paging
- https://devblogs.microsoft.com/odata/using-skiptoken-for-paging-in-asp-net-odata-and-asp-net-core-odata/

I could focus only on `$skiptoken`, but I felt it would be incomplete. And given the common issues we get where customers confuse the roles of `PageSize` and `$top`, I thought it would be beneficial to dedicate a section on that. And given that `$skiptoken` is an improvement, I thought it would make more sense to "lead into it". I also considered that someone may come to this page without having read the client-driven paging docs. Open to suggestions.

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: Would that not make the sentence too long?

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
habbes: I rephrased it based on the suggestion.

Odata-docs/TOC.yml
habbes: Should I also swap "Pagination => Paging" in the client docs TOC? Not sure if it might lead to dead links.

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: There's no standard way to handle the skip token, as this is up to the server's implementation. The default skip token handler in WebAPI 8 handles this conceptually similar to the example I showed using `GET /Products?$filter=Id gt 2&$top=2`. I talk more about the `DefaultSkipTokenHanlder` in a separate [PR](https://github.com/MicrosoftDocs/OData-docs/pull/283) that shows how to customize the `$skiptoken` logic. I can also briefly mention in this doc.

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
habbes: See: https://github.com/MicrosoftDocs/OData-docs/pull/281#discussion_r1087443168

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
gathogojr: @habbes It's because "By default, the built-in DefaultSkipTokenHandler class." doesn't feel like a complete sentence

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
gathogojr: @habbes Our offline conversation refers. It was a knowledge gap on my part. I'm happy with the document covering both `$skip` and `$skiptoken`

Odata-docs/TOC.yml
gathogojr: I'd say only on the changes in this PR...

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: I have added a brief explanation of how the default skip token handler handles the skip token and why the generated query is more efficient than using `$skip`.

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
odero: By default, the built-in [`DefaultSkipTokenHandler`](/dotnet/api/microsoft.aspnetcore.odata.query.defaultskiptokenhandler) class generates the `$skiptoken` based on the values of the key fields and fields in the `$orderby` query option if present

What about this?

habbes: Aah I see, my bad.

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,268 @@
...
+            "Name": "Product 2" 
+        },
+    ],
+    "@odata.nextLink": "http://localhost:5000/odata/Products?$skiptoken=foobar"
+}
+```
+
+In this example, the client would make a request to `GET http://localhost:5000/odata/Products?$skiptoken=foobar` to retrieve the next page.
+
+The format of the next link is opaque to the client. This means that the client should not try to decode the link, modify it or make any assumptions about how it is generated or what its components mean. It can only assume that it's a valid URL that will return the next page. This allows the server freedom in how it chooses to generate and interpret the next link. It allows the server to change the underlying implementation of the next link without affecting the client.
```
gathogojr, line 37:
```suggestion
The format of the next link is opaque to the client. This means that the client should not try to decode the link, modify it or make any assumptions about how it is generated or what its components mean. It can only assume that it's a valid URL that will return the next page. This allows the server freedom in how it chooses to generate and interpret the next link. It also allows the server to change the underlying implementation of the next link without affecting the client.
```

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: Rephrased to "By default, the built-in `DefaultSkipTokenHandler` class generates..."

Odata-docs/TOC.yml
odero: @habbes Did you check if this change would result in dead links or not?

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,187 @@
...
+```csharp
+public class ProductsController : ODataController
+{
+    [EnableQuery]
+    public IQueryable<Product> Get()
+    {
+        return productsCollection;
+    }
+}
+```
```
gathogojr, line 63: @habbes I don't feel very strongly about this but based on how you have presented this section, I feel that it's a bit hard for someone to know that configuring client-side paging involves 2 steps.
I feel like a reorg of the section could help, starting with a heading level 2 "Configuring client-side paging | Configuring client-side paging in OData".
Then, the next line could be "Configuring client-side paging involves two steps:". Then a numbered list comprising of first, configuring the relevant OData services/features and second about the `EnableQuery` attribute. That way, it'd be clear that the two steps are required.
Some people have a tendency of trying things out as they read and without the indication that it's a two-step process, they could get stuck wondering why it doesn't work after performing step 1 only.

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Benefits of this approach
```
gathogojr, line 65: Maybe "Benefits of client-side paging approach" or "Advantages of client-side paging approach"?

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,187 @@
...
+
+This client-driven paging approach has a number of benefits:
+
+- It's relatively easy to implement
+- It allows the client to fetch pages independently and to jump to an arbitrary page. For example, the client can fetch page 5 without having fetched page 4 previously.
+- If the client knows the total size of the collection (e.g. using the `$count` query option), the client can compute the total number of pages as well as generate links for each page in advance.
+
+## Drawbacks
```
gathogojr, line 73: Maybe "Drawbacks of client-side paging approach" or "Disadvantages of client-side paging approach"

gathogojr, line 73: You used an unordered list in the "Benefits" section but you adopted prose here... Was that because there's only one drawback or because it's easier to talk about it in prose form? Does all the content that follow this section fall under "Drawbacks"? Is there a way you could have broken down the section in an ordered/unordered list?

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,187 @@
...
+        return productsCollection;
+    }
+}
+```
+
+## Benefits of this approach
+
+This client-driven paging approach has a number of benefits:
+
+- It's relatively easy to implement
```
KenitoInc, line 69:
```suggestion
- It's relatively easy to implement.
```

Odata-docs/TOC.yml
habbes: The build output should report any broken links (for relative paths). I've also tested the paging links and haven't found a broken link so far.

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
habbes: Good observation. I think opted for a prose format because I explain the drawbacks with detailed examples as opposed to just listing them. I think I mention 2 main drawbacks in this section and the end the section in a way that alludes to server-side paging. I could start by listing the disadvantages, using an ordered list, then follow the list with a detailed explanation for each.

habbes: I have re-organized this section to start with a list, then illustrate the issues in more detail. Not sure if that's any better though.

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,197 @@
...
+        "code": "",
+        "message": "The query specified in the URI is not valid. The limit of '0' for Top query has been exceeded. The value from the incoming request is '2'."
+}
+```
+
+## Configuring client-side paging in OData
+
+Configuring client-side paging involves two steps:
+
+1. Configuring the max `$top` query option value
```
ElizabethOkerio, line 36:
```suggestion
1. Configuring the `$top` query option with a max value
```

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
```diff
@@ -0,0 +1,197 @@
...
+```csharp
+services.AddOData(options =>
+    options.SetMaxTop(100)
+    //...
+    );
+```
+
+Now this will support values of up to 100, but return an error if `$top` exceeds 100.
+
+You can also remove the limit by setting it to `null`:
```
ElizabethOkerio, line 50: does this mean when you set it to null you can use $top with whichever value and it will work.. I think you should say that explicitly if that's the case

Odata-docs/webapi-8/fundamentals/paging-overview.md
```diff
@@ -0,0 +1,21 @@
...
+ms.author: clhabins
+---
+
+# Paging in ASP.NET Core OData 8
+
+**Applies To**:[!INCLUDE[appliesto-webapi](../../includes/appliesto-webapi-v8.md)]
+
+Loading large datasets can be slow. Services and clients often rely on paging to load data in chunks or pages to improve the response times and user experience. ASP.NET Core OData provides support for client-driven and server-driven paging models.
+
+In client-driven paging, the client uses the `$top` and `$skip` query options to request a specific page of content from the server. In server-driven paging, the server limits the number of items per response page and includes a **next link** in the response that the client can use to request the next page.
```
ElizabethOkerio, line 16: nextlink ~ this should be a one word and not two words

ElizabethOkerio: if you agree with this then update in all the other places.

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
```diff
@@ -0,0 +1,268 @@
...
+ms.date: 1/11/2023
+author: habbes
+ms.author: clhabins
+---
+
+# Server-driven Paging in ASP.NET Core OData 8
+
+**Applies To**:[!INCLUDE[appliesto-webapi](../../includes/appliesto-webapi-v8.md)]
+
+In [server-driven paging](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_ServerDrivenPaging), the server returns the first page of results. If total number of results is greater than the page size, the server returns the first page along with a "next link" that can be used to fetch the next page of results. Each subsequent page should include a next link except for the last page. Using this approach, the client needs to fetch pages sequentially. It cannot jump to an arbitrary page. The next link is included in the response using the [`@odata.nextLink`](http://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_ControlInformationnextLinkodatanextL) annotation.
```
ElizabethOkerio, line 14: you've provided a link to the v4.01 format but you are using a v4.0 format.

Odata-docs/webapi-8/fundamentals/paging-overview.md
habbes: I was referring to next link not as a keyword, but words. Where I used as a keyword I formatted it as code, e.g. `nextLink` or `@odata.nextLink`.

habbes: Similar to how I used skip token vs`$skiptoken`

Odata-docs/webapi-8/fundamentals/client-driven-paging.md
habbes: It means there's no (upper) limit. Any `$top` value >= 0 will be allowed.

Odata-docs/webapi-8/fundamentals/server-driven-paging.md
habbes: Thanks. Updated.

PR #296, opened 2023-05-22
Update getting-started.md

michelcedric:
Adapt get method by key

An odata endpoint must return a IQueryable.
If you evaluate it directly in the return of the controller method, the odata query options can't working.
If you directly return the queryable you return an array with only one element.
You can fix that with the usage of 
SingleResult.Create method form odata

### DIFF
--- a/Odata-docs/webapi-8/getting-started.md
+++ b/Odata-docs/webapi-8/getting-started.md
@@ -150,14 +150,8 @@ namespace Lab01.Controllers
         [EnableQuery]
         public ActionResult<Customer> Get([FromRoute] int key)
         {
-            var item = customers.SingleOrDefault(d => d.Id.Equals(key));
-
-            if (item == null)
-            {
-                return NotFound();
-            }
-
-            return Ok(item);
+            var item = customers.AsQueryable().Where(c=>c.Id==key);           
+            return SingleResult.Create(item);
         }
     }
 }


michelcedric:
@microsoft-github-policy-service agree

Odata-docs/webapi-8/getting-started.md
```diff
@@ -148,16 +148,10 @@ namespace Lab01.Controllers
         }
 
         [EnableQuery]
-        public ActionResult<Customer> Get([FromRoute] int key)
```
gathogojr, line 151: There's no reason for this change. `$expand` works okay even when the return type is `ActionResult<Customer>`.

Odata-docs/webapi-8/tutorials/basic-crud.md
```diff
@@ -376,20 +376,14 @@ The following JSON payload shows the expected response:
 ## Request a single entity
 To support this request, we add a controller action named `Get` (or `GetCustomer`) to the `CustomersController` class. The action should accept a single parameter named `key` of type `int` - same type as the entity's key property:
 ```csharp
-public ActionResult<Customer> Get([FromRoute] int key)
```
gathogojr, line 379: There's no reason for this change. `$expand` works okay even when the return type is `ActionResult<Customer>`.

Odata-docs/webapi-8/tutorials/basic-crud.md
```diff
@@ -398,14 +392,8 @@ Note that for the built-in routing conventions to route the request successfully
...
-    var customer = db.Customers.SingleOrDefault(d => d.Id == id);
-
-    if (customer == null)
-    {
-        return NotFound();
-    }
-
-    return Ok(customer);
+    var item = customers.AsQueryable().Where(c => c.Id == id);
+    return SingleResult.Create(item);
```
gathogojr, line 396: There's no reason for this change. `$expand` works okay even when the return type is `ActionResult`.

gathogojr:
@michelcedric There's no reason for the changes that you proposed. Query options work okay even when the return type is `ActionResult` or `ActionResult<T>`. If you have reasons to believe that they don't, please report the issue together with a simple repro on https://github.com/OData/AspNetCoreOData/issues.

michelcedric:
@gathogojr 
I'm sorry, I will give more details.
But If you use a real Database as Sql Server with EF.
When you execute.
`var customer = db.Customers.SingleOrDefault(d => d.Id == key);`
The query is immediately evaluated and executed on the database.
Customer object is loaded in memory and in this case expand is not possible.
It's that why the return of the controller must be a queryable to delay to Odata the execution of the query.
For an id the return is not a queryable but only the entity. It's seems the reasons why SingleResult of Odata exists.
` var item = customers.AsQueryable().Where(c => c.Id == id);
  return SingleResult.Create(item);`
I created a repository to exactly show what I mean
https://github.com/michelcedric/OdataInclude/blob/main/OdataInclude/Controllers/BooksController.cs
https://github.com/michelcedric/OdataInclude/blob/main/OdataInclude/OdataInclude.http
image
1 Get books to retrieve the first ID
2 Get book with the ID and expand MainAuthor=> MainAuthor null
3 Get book with the ID and expand MainAuthor version 2 => MainAuthor filled
4 Get book with bad ID and expand MainAuthor=> Not found 404.

gathogojr:
> @gathogojr I'm sorry, I will give more details. But If you use a real Database as Sql Server with EF. When you execute. `var customer = db.Customers.SingleOrDefault(d => d.Id == key);` The query is immediately evaluated and executed on the database. Customer object is loaded in memory and in this case expand is not possible. It's that why the return of the controller must be a queryable to delay to Odata the execution of the query. For an id the return is not a queryable but only the entity. It's seems the reasons why SingleResult of Odata exists. ` var item = customers.AsQueryable().Where(c => c.Id == id); return SingleResult.Create(item);` I created a repository to exactly show what I mean https://github.com/michelcedric/OdataInclude/blob/main/OdataInclude/Controllers/BooksController.cs https://github.com/michelcedric/OdataInclude/blob/main/OdataInclude/OdataInclude.http image 1 Get books to retrieve the first ID 2 Get book with the ID and expand MainAuthor=> MainAuthor null 3 Get book with the ID and expand MainAuthor version 2 => MainAuthor filled 4 Get book with bad ID and expand MainAuthor=> Not found 404.

@michelcedric The sample code in this tutorial uses an in-memory collection as a data store.
There's therefore no value in returning a `SingleResult` in this case.
There's also no value in casting the collection to `AsQueryable`.

In my opinion, it'd be more ideal to have a different article with code samples that are relevant to an SQL Server data store where we can employ the use of `SingleResult.Create` to construct and execute the query for the expanded entity at the database level...

PR #339, opened 2024-11-09
Update basic-crud.md

montoner0:
Added an important note about controller naming

### DIFF
--- a/Odata-docs/webapi-8/tutorials/basic-crud.md
+++ b/Odata-docs/webapi-8/tutorials/basic-crud.md
@@ -312,7 +312,7 @@ In the controller, we add a contructor that accepts a `BasicCrudDbContext` param
 We will add controller actions to this controller progressively to support different CRUD operations.
 
 > [!IMPORTANT]
-> If you've made changes to the code, make sure that the controller's name matches the entity set's name registered in the entity data model (e. g. modelBuilder.EntitySet<Customer>("**Customers**") matches **Customers**Controller).
+> If you've made changes to the code, make sure that the controller's name matches the entity set's name registered in the entity data model (e. g. modelBuilder.EntitySet\<Customer>("**Customers**") matches **Customers**Controller).
 > Failing to do so will result in the controller's endpoints aren't registered properly.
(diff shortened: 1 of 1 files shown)


gathogojr:
@montoner0 The rules surrounding naming of controllers (and controller actions) when using conventional routing is already covered in the Fundamentals https://learn.microsoft.com/en-us/odata/webapi-8/fundamentals/routing-overview?tabs=net60#odata-routing.

I don't believe we need to repeat it here as well. It is expected that someone understands the fundamentals first before trying out the tutorials.

Odata-docs/webapi-8/tutorials/basic-crud.md
```diff
@@ -311,6 +311,12 @@ In the controller, we add a contructor that accepts a `BasicCrudDbContext` param
 
 We will add controller actions to this controller progressively to support different CRUD operations.
 
+> [!IMPORTANT]
+> If you've made changes to the code, make sure that the controller's name matches the entity set's name registered in the entity data model (e. g. modelBuilder.EntitySet\<Customer>("**Customers**") matches **Customers**Controller).
```
WanjohiSammy, line 315: I think it’s helpful to know that a controller’s name should match the entity set’s name registered in the entity data model. However, this isn’t essential because the controller name can be overridden using the `[Route("api/Items")]` attribute.

For example, you can have the following controller:
```cs
[Route("api/Items")]
[ApiController]
public class HandleAllController : ODataController
{
    // Controller actions go here
}
```

montoner0: > the controller name can be overridden using the [Route("api/Items")] attribute

Agree, but in this case you need to specify a route for every endpoint with parameters, otherwise they won't work.

WanjohiSammy: Can you also add the other option of specifying a route for every endpoint.

montoner0: Sorry, I don't quite get what you want.

PR #346, opened 2025-02-19
IODataClientFactory.CreateClient<T> requires Uri parameter

woha:
The provided code example does not work.

IODataClientFactory.CreateClient<T> requires an Uri parameter in addition to the client name.

### DIFF
--- a/Odata-docs/client/using-extensions.md
+++ b/Odata-docs/client/using-extensions.md
@@ -73,8 +73,6 @@ If an app requires many distinct uses of OData client proxy, each with a differe
 services.AddODataClient("TripPin")
     .ConfigureODataClient(dsc =>
     {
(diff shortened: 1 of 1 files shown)


Odata-docs/client/using-extensions.md
```diff
@@ -99,7 +97,8 @@ public class PeopleController : ODataController
     [EnableQuery]
     public IEnumerable<Person> Get()
     {
-        var client = _clientFactory.CreateClient<DefaultContainer>("TripPin");
+        var baseUri = new Uri("https://services.odata.org/v4/(S(lqbvtwide0ngdev54adgc0lu))/TripPinServiceRW/");
+        var client = _clientFactory.CreateClient<DefaultContainer>(baseUri, "TripPin");
```
WanjohiSammy, line 101: Your suggestion is to do this for every action

